PR #89, opened 2026-04-16
Potential fix for code scanning alert no. 2: Prototype-polluting function

hamidfzm:
Potential fix for [https://github.com/hamidfzm/glyph/security/code-scanning/2](https://github.com/hamidfzm/glyph/security/code-scanning/2)

General fix: constrain deep assignment so recursive traversal and final write are only allowed through known-safe, own-property paths on the destination schema, rather than accepting arbitrary path segments.

Best concrete fix in `src/contexts/SettingsContext.tsx`: in `setNestedValue`, before creating/interacting with nested objects, verify each segment is an own property of the corresponding node in `DEFAULT_SETTINGS` (schema object). Keep existing forbidden-key checks, but add schema-walk validation for both intermediate keys and final key. If any segment is invalid, return original `obj`. This preserves behavior for legitimate settings paths while blocking any unexpected or attacker-crafted chains.

Changes needed:
- In `setNestedValue`, introduce a `schemaCurrent` pointer initialized from `DEFAULT_SETTINGS`.
- During loop (`keys.length - 1`), require `Object.hasOwn(schemaCurrent, key)` and advance `schemaCurrent`.
- Before final assignment, require `Object.hasOwn(schemaCurrent, lastKey)`.
- No new imports/dependencies needed.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/src/contexts/SettingsContext.tsx
+++ b/src/contexts/SettingsContext.tsx
@@ -66,15 +66,24 @@ function setNestedValue(
 
   const result = { ...obj };
   let current: Record<string, unknown> = result;
+  let schemaCurrent: Record<string, unknown> = DEFAULT_SETTINGS as unknown as Record<string, unknown>;
 
   for (let i = 0; i < keys.length - 1; i++) {
     const key = keys[i];
     if (FORBIDDEN_OBJECT_KEYS.has(key)) {
       return obj;
     }
-    if (!isSafePlainObject(current)) {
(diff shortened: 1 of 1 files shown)


hamidfzm:
Superseded by #91, which applies the same schema-allowlist logic with correct formatting, conventional commit title, current dompurify override from main, and regression tests that fail without the fix.

PR #328, opened 2026-06-22
feat(i18n): Persian translation and RTL layout support

### LINKED ISSUE #127: Add internationalization (i18n) and community translations
## Problem

Glyph is English-only. Every user-facing string — menu items, settings labels, modals, AI panel copy, error toasts, empty states — is hard-coded. This caps adoption to English speakers and locks out the easiest form of community contribution: translation.

Most comparable apps ship multi-language UIs (Obsidian: 30+ locales; Joplin: 50+; Typora: ~20).

## Proposed Solution

i18n plumbing in the app, with translations kept **in this repo** (monorepo). No separate translations repo: locale JSON files live under `src/locales/<lng>/`, and community translators contribute via PRs to this repo.

- Runtime: **react-i18next** (JSON files translators edit directly; native Crowdin/Weblate support if we adopt one later; English inline as the always-available fallback).
- Wrap user-facing strings with `t("key")` calls, namespaced by surface (`common`, `settings`, `menu`, `ai`) so bundles can lazy-load. Largest surfaces:
  - `src-tauri/src/menu.rs` / `menu_runtime.rs` — File/Edit/View/Help labels. The menu is built in Rust, so `build_menu` takes a localized label struct and is **rebuilt on locale change** via a Tauri command.
  - `src/components/modals/settings/*`, `AIPanel.tsx` — biggest cluster of strings
  - `src/components/layout/EmptyState.tsx`, `Sidebar.tsx`, `StatusBar.tsx`, `Titlebar.tsx`
  - Error/toast strings throughout hooks
- **Locale detection (OS-level, not webview):** resolve the active locale in this order — explicit user override (`appearance.locale` ≠ `system`) → OS UI language via `@tauri-apps/plugin-os` `locale()` → `navigator.language` (fallback) → `en`. Walk the BCP-47 fallback chain (`pt-BR → pt → en`). `navigator.language` alone is wrong here: in a Tauri webview it reflects the embedded browser locale, not the desktop UI language.
- Settings UI: a language picker. Locales with ≥80% completion show normally; others stay available but flagged.
- English source of truth lives at `src/locales/en/*.json`; other locales are sibling folders in the same repo.

## Build order

1. **Foundation:** runtime (`src/lib/i18n.ts`), locale registry (`src/lib/locales.ts`), `appearance.locale` setting, `useLocale` hook (OS detection, sets `<html lang>`/`dir`), `en` namespaces, language picker. Ship one fully-translated locale (`de`) as an end-to-end smoke test.
2. **Migrate React strings** to `t()`, surface by surface.
3. **Native menu** localization (rebuild-on-locale-change in Rust).
4. Docs + remaining seed locales.

## Initial languages to seed

Tracked in #265 (LTR-first tiers). RTL locales (Arabic, Hebrew, Persian) are split into #264 since they need a layout audit, not just translation.

## Risks

- **Native menu localization** is more work than React strings; menus are built in Rust at app start. Rebuild the menu when locale changes (chosen) over passing strings once at startup.
- **Translation rot** — strings change faster than translators can keep up. Mitigate with runtime English fallback for missing keys, and (optionally) a schema-validation CI step that fails on missing keys against `en`.
- **Bundle size** — eagerly loading all locales bloats startup. Lazy-load non-`en` locales; keep `en` inline as the fallback.

## Alternatives Considered

- **Stay English-only** — lowest cost; loses non-English markets entirely.
- **Separate translations repo + nightly sync Action** — was the original plan; dropped in favor of a monorepo. Keeping locales in-repo means one source of truth, normal PR review, and no sync infrastructure to maintain.
- **Single big locales JSON** instead of per-namespace — works for our size today; harder to lazy-load by surface later.

## Platform

All

## Related

- RTL layout support — #264
- Seed languages tracking — #265
- Marketing website (#115) — should be localized at the same time
- Plugin / extension API — the runtime should expose `t()` to plugins eventually

### LINKED ISSUE #264: RTL (right-to-left) layout support for Arabic, Hebrew, and Persian
## Problem

Glyph's UI is built with physical CSS directions (`margin-left`, `padding-right`, `left:`/`right:`, hard-coded sidebar sides) and a fixed `<html dir="ltr">`. Once the i18n work (#127) lands, RTL locales (Arabic, Hebrew, Persian/Farsi) will have translated strings but a broken left-to-right layout: the sidebar, scrollbars, drag regions, code-block gutters, tab overflow, and command palette will all read the wrong way.

This is split out of #127 (i18n) so the LTR translation work can ship first; RTL is a focused layout audit that depends on the locale plumbing but is otherwise independent. It also overlaps with the accessibility work in #134 (correct `dir`/`lang` are WCAG requirements).

## Proposed Solution

Make the layout direction-aware and driven by the active locale.

- **Direction from locale.** Extend the locale registry/`useLocale` hook (from #127) to expose a `dir` per locale and set `<html dir="rtl">` for `ar`, `he`, `fa`. Mirrors how theme/`lang` are applied today.
- **CSS logical properties.** Convert physical properties in `src/styles/*` (~21 occurrences of `margin-left/right`, `padding-left/right`, `left:`/`right:`) to logical equivalents: `margin-inline-start/end`, `padding-inline-*`, `inset-inline-*`, `text-align: start/end`.
- **Layout audit** of the surfaces called out in #127's Risks:
  - Sidebar position and the existing `swapSidebarSides` logic
  - Scroll direction, resize handles, and titlebar drag regions
  - Code-block margins/gutters, inline-code padding
  - Command palette, tab bar overflow, status bar alignment
  - Icon mirroring where directional (chevrons, back/forward, indentation guides)
- **Tests.** Render representative components with `dir="rtl"` and assert logical layout; add an RTL sample under `samples/` for manual QA.
- Pairs well with the Claude design review (#114).

## Alternatives Considered

- **Bundle RTL into #127** — makes that PR much larger and delays shipping LTR locales; the issue itself recommends deferring RTL.
- **Per-component JSX conditionals on direction** — clutters components; CSS logical properties handle the bulk automatically.

## Platform

All

## Additional Context

Depends on #127 (i18n plumbing: locale setting, `useLocale`, `<html lang>`). Related: #134 (WCAG 2.2 AA), #114 (design review).

hamidfzm:
## Summary

Follow-up to #325. The auto-merge on #325 fired on the English string-extraction scope before the Persian and RTL commits were pushed, so those two never reached `main` (the squashed commit `a7182c5` carries only the `en` extraction). This PR brings them in:

1. **Persian translation**: a complete `fa` bundle across all six namespaces (402 keys, exact key parity with `en`, verified, so nothing falls back to English). Selectable in Settings, Appearance.
2. **RTL layout support**: the UI is now direction-aware so Persian mirrors correctly.

## RTL details

- Physical CSS converted to logical properties across chrome and content styles (`margin/padding/border-inline-*`, `inset-inline-*`, `text-align: start/end`).
- Directional Tailwind utilities converted to logical (`ms/me/ps/pe`, `start/end`, `border-s/e`).
- `dir`-aware `transform` overrides for the settings toggle thumb and the AI panel slide-in (their `translateX` is physical).
- The sidebar mirrors via the existing flex-row reversal; panel/edge dividers face content through logical borders.
- Deliberately NOT mirrored: the canvas board and graph are coordinate-based spatial surfaces (matches Obsidian); centering pairs (`left:50%` + `translateX(-50%)`) and the macOS traffic-light spacer stay physical.
- Also moves a hardcoded `"Reset view"` graph label (missed by the string sweep) into `graph.resetView` (en + fa).

## Testing

- `pnpm typecheck`, `pnpm check` (Biome), and `pnpm test` pass.
- Tests for locale resolution, `fa` to `dir="rtl"`, the picker option, plus a key-parity check (en vs fa).
- Manual: Settings, Appearance, choose فارسی; the whole UI switches to Persian and flips RTL.

Closes #264.
Part of #127.


hamidfzm:
Note on the native menu: its **text** is localized (via `set_menu_labels`), but the menu bar **order does not mirror under RTL** — Tauri's `muda` exposes no direction/RTL API, and `WS_EX_LAYOUTRTL` would double-mirror the webview. Full RTL for the menu would require a custom in-app menu bar, tracked in #335.

PR #469, opened 2026-07-16
feat(editor): multi-language spell check with script-aware skipping

### LINKED ISSUE #426: Support multiple spell-check languages and stop flagging scripts with no enabled dictionary
## Problem

Spell check assumes one active language, which breaks on mixed-language documents. With English active, every Persian word in an English/Persian note is underlined as misspelled (and vice versa): the single active dictionary rejects everything outside its own language. Bilingual notes are the norm for Persian users, so the feature is effectively unusable for them today. There is also no way to enable two dictionaries at once even when both are installed.

## Proposed Solution

Check against a set of enabled dictionaries instead of exactly one.

- Settings: replace `editor.spellCheckLanguage: string` with `editor.spellCheckLanguages: string[]` (default `["en"]`), migrating the old persisted key on load in the same PR. The Settings picker becomes a checkbox list of built-in English plus every installed dictionary.
- Checking: a word is correct if any enabled dictionary accepts it; it is flagged only when all enabled dictionaries reject it.
- Script awareness: a word whose script no enabled dictionary covers is skipped, not flagged. With only English enabled, Persian text stops lighting up entirely; enabling the Persian dictionary starts checking it. Each dictionary declares (or the host infers from its language code) the scripts it covers.
- Suggestions: merged from the dictionaries that cover the word's script, deduplicated, capped as today. Add to dictionary adds to the matching dictionaries' session set; Ignore stays as is.
- The speller cache and the CodeMirror extension move from one language to a language set; toggling any language reconfigures in place (Compartment), preserving cursor and undo.

## Acceptance Criteria

- [x] Settings offers a multi-select of installed dictionaries; the stored value is a string array with the old single value migrated
- [x] A word accepted by any enabled dictionary is not underlined
- [x] A word in a script covered by no enabled dictionary is not underlined (mixed English/Persian text with only English enabled shows no Persian false positives)
- [x] A word rejected by all enabled dictionaries that cover its script is underlined
- [x] Right-click suggestions come from the dictionaries covering the word's script; Ignore and Add to dictionary behave per current semantics
- [x] Enabling or disabling a language updates underlines without recreating the editor (cursor and undo intact)
- [x] Vitest covers the union check, script skipping, migration, and the multi-select UI

## Scope / Out of scope

**In scope:** multi-dictionary checking, script-based skipping, settings migration, multi-select UI, suggestion merging.

**Out of scope:** per-document language hints from frontmatter `lang:` (follow-up if wanted), persisting personal words to disk, sandboxed-plugin dictionary support.

## Implementation Tasks

- [x] Settings shape + migration + multi-select UI
- [x] Script coverage map + word-script detection in the scanner
- [x] Union check in the spellcheck extension; suggestion merging in the menu
- [x] Speller-set plumbing (cache keyed per language, extension takes a set)
- [x] Tests + coverage
- [x] Docs (README bullet wording)

## Affected areas

Settings, Modals (settings), Markdown rendering (editor)

## Platform

All

hamidfzm:
## Summary

Spell check now runs against a set of enabled dictionaries instead of exactly one, and words are only checked against dictionaries covering their script. This fixes mixed-language documents: with only English enabled, Persian text is no longer flagged wholesale (it is skipped), and enabling English + Persian together checks each word against the right dictionary and flags it only when every covering dictionary rejects it.

Closes #426

## Changes

- `editor.spellCheckLanguage` (single) becomes `editor.spellCheckLanguages` (array, default `["en"]`), with an idempotent migration in `settingsMigrations.ts` seeding the array from the legacy key and dropping it
- New `src/lib/spellcheck/scripts.ts`: classifies a word's script from its first letter (`\p{Script=...}`) and maps language codes to primary scripts; `DictionaryContribution` gains an optional `scripts` field for dictionaries whose coverage the code cannot infer
- `spellcheckExtension.ts`: checkers set with union accept / unanimous reject / uncovered-script skip; dictionaries load independently and repaint progressively; suggestions merge deduped across covering dictionaries; Add to dictionary routes to covering dictionaries; the session ignore list is global (was per language)
- Settings language picker becomes toggle rows (built-in English + installed dictionaries); an orphaned enabled code stays listed so it can be turned off; empty set checks nothing
- Compartment reconfigure keys on the joined language set, so settings-save array identity churn does not touch the editor
- Locale keys updated in all five languages (en, de, es, fa, zh)

**Deliberate non-bump:** the plugin API stays at 0.16.0. The new `scripts` contribution field is optional with a safe inferred default, and under the exact-match 0.x regime (#416) a bump would reject every installed plugin over an additive field. Flagging for maintainer override if this counts as a "decision set" change.

## Testing

- [ ] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux
- [x] `pnpm typecheck`, `pnpm check`, `pnpm test` (2567 passing) green; `cargo clippy` clean
- [x] Patch coverage verified locally from lcov: every changed file has no uncovered lines from this diff

Tests cover: union accept / unanimous reject / script skip (unit + mounted editor), suggestion merging and dedupe, Add routing by script, progressive load with a pending dictionary, global ignore across editors, the settings migration (seed, idempotency, composition), script classification incl. ZWNJ-leading Persian words, toggle-row set math incl. orphaned codes, and Compartment reconfigure on real set changes only.

## Screenshots

Settings → Editor now lists per-language toggles under Spell Check; mixed English/Persian text shows underlines only for languages you enabled.


hamidfzm:
Follow-up commit 35f6072 (from self-review): script coverage is now general purpose instead of enumerated. Language codes resolve through Intl.Locale likely-subtags (POSIX-style fa_IR accepted) and coverage uses dynamic \p{Script=...} probes, so every script Unicode knows works with no hand-maintained tables. The scriptCoverage seam canonicalizes casing, expands ISO 15924 composites (Jpan/Kore/Hans/Hant) for plugin-declared lists too, rejects crafted names that would break out of the probe pattern, and warns on unknown names instead of silently covering nothing. Review also hardened teardown (a replaced configuration can no longer paint stale marks from a late dictionary load or an orphaned menu action) and the migration now discards a corrupt non-array spellCheckLanguages instead of letting the editor crash on mount. All files remain at 100 percent patch coverage locally.

Issue #290, opened 2026-06-11
Publish Glyph to Flathub (Flatpak)


hamidfzm:
## Problem

Flathub is the largest discovery + install channel for Linux desktop apps, and the one most users reach for first. Glyph currently ships to Linux via AUR, PPA, the apt repo, Homebrew, and raw `.deb`/`.AppImage`, but it has no Flatpak. That leaves out every user on Fedora, the Steam Deck, elementary OS, Pop!_OS GNOME Software, KDE Discover, and any distro where Flatpak is the default app source.

## Proposed Solution

Publish Glyph to [Flathub](https://flathub.org/) as `com.hamidfzm.glyph` (matches the existing Tauri `identifier`).

In-repo scaffolding (this repo):

- [x] Add `flatpak/com.hamidfzm.glyph.metainfo.xml` (AppStream metainfo: name, summary, description, screenshots, release notes, content rating, branding). Flathub quality-gates on this file.
- [x] Add `flatpak/com.hamidfzm.glyph.desktop` (desktop entry: `Exec`, `Icon`, categories, MIME types for markdown so Glyph shows up as a handler).
- [x] Add `flatpak/com.hamidfzm.glyph.yml` manifest that installs the released `.deb` into the `org.gnome.Platform` (GNOME 50) runtime, following the [Tauri Flatpak guide](https://v2.tauri.app/distribute/flatpak/). Per-arch deb sha256s pinned for v0.10.0 and verified.
- [x] Wire CI validation (`validate-packaging`: `appstreamcli validate`, `desktop-file-validate`, YAML parse) so the metainfo/desktop/manifest stay valid.

Submission + publish (one-time + ongoing):

- [ ] Submit the manifest as a PR to [flathub/flathub](https://github.com/flathub/flathub) `new-pr` branch. A first attempt (flathub/flathub#9004) was closed/rejected because Flathub's generative-AI policy disallows AI-opened submission PRs and AI-generated manifests/prose. If pursuing Flathub, the maintainer must author and open the submission by hand.
- [x] Update strategy decided: the canonical manifest stays in this repo; the `publish-flathub` job in `release.yml` pushes each release to `flathub/com.hamidfzm.glyph` (its buildbot then publishes).
- [x] Add Flathub install instructions to `README.md` and the release-notes `Install` block in `release.yml`.

### Maintainer action items (only the repo owner can do these)

[...]

Issue #117, opened 2026-05-01
Submit Glyph to awesome-markdown-editors


hamidfzm:
## Problem

Glyph isn't listed on https://github.com/mundimark/awesome-markdown-editors, which is a major discovery channel for markdown app users. Comparable apps (Inkwell, Markpad, MarkText, Ferrite, Kindling) are all listed.

### Submit a PR to awesome-markdown-editors

> ✅ **Submitted:** mundimark/awesome-markdown-editors#169 (added to `UPCOMING.md` → Markdown Desktop Editors → Universal, per the 2026 upcoming-first policy).

1. Read the contribution guidelines in that repo's README / CONTRIBUTING.
2. Add Glyph under the appropriate section. Likely "Universal (Cross-Platform)" given macOS/Windows/Linux support.
3. Suggested entry:

   > **Glyph** (FREE, open source): Native markdown viewer and editor built with Rust/Tauri. Features GitHub Flavored Markdown, KaTeX math, Mermaid diagrams, multi-file tabs, AI summarization (Claude/OpenAI/Ollama), text-to-speech, and platform-native styling. Cross-platform: macOS, Windows, Linux.

4. Link to the GitHub repo and (once it exists) the marketing website.

### Other lists & directories to submit to

**Awesome lists (submitted):**

- [x] [mundimark/awesome-markdown-editors](https://github.com/mundimark/awesome-markdown-editors) - ✅ merged: mundimark/awesome-markdown-editors#169 (`UPCOMING.md` → Desktop → Universal)
- [x] ~[tauri-apps/awesome-tauri](https://github.com/tauri-apps/awesome-tauri) - submitted: tauri-apps/awesome-tauri#730 (Office & Writing)~
- [x] [jaywcjlove/awesome-mac](https://github.com/jaywcjlove/awesome-mac) - submitted: jaywcjlove/awesome-mac#2146 (Markdown Tools)
- [x] [tehtbl/awesome-note-taking](https://github.com/tehtbl/awesome-note-taking) - submitted: tehtbl/awesome-note-taking#86 (Open Source → Tauri)
- [x] [0PandaDEV/awesome-windows](https://github.com/0PandaDEV/awesome-windows) - submitted: 0PandaDEV/awesome-windows#189 (Text Editors)

**Awesome lists (blocked / later):**

[...]

PR #514, opened 2026-07-19
docs(packaging): advertise Flathub install and smoke-test the channel

hamidfzm:
## Summary

Second half of the Flathub rollout, split from #467. That PR lands the inert packaging (manifest, metadata, validate job, token-gated publish job); this one flips the user-facing switch. **Keep as draft until the app is live on Flathub**, otherwise the README and release notes would point at a package that does not exist yet.

Based on `feat/flathub` so the diff stays scoped; GitHub retargets it to `main` automatically when #467 merges.

## Changes

- README: `### Linux (Flathub)` install section
- `release.yml`: Flathub install lines in both Linux release-notes blocks
- `release-smoke-test.yml`: new `flathub` channel job (version check against the Flathub remote, install, headless launch under xvfb, `pending`/`verified` status) wired into the summary table, modeled on the existing `snap` job

## Testing

- [ ] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

YAML for both workflows parses; the smoke-test job reports `pending` until Flathub serves the released version, so merging it before the first Flathub build only adds a harmless pending row.

Refs #117, #290

### DIFF
--- a/.github/workflows/release-smoke-test.yml
+++ b/.github/workflows/release-smoke-test.yml
@@ -278,6 +278,37 @@ jobs:
           [ "$CODE" -eq 124 ] || { echo "launch check failed (exit $CODE)"; exit 1; }
           echo "status=verified" >> "$GITHUB_OUTPUT"
 
+  flathub:
+    needs: resolve
+    if: needs.resolve.outputs.run == 'true'
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


hamidfzm:
Not perusing this because of flathub anit community policies

PR #630, opened 2026-08-12
perf(bundle): enforce bundle budgets and trim startup by 15%

### LINKED ISSUE #437: Reduce startup and optional-feature bundle costs with enforced budgets
## Problem

The production build succeeds but reports multiple oversized chunks. Audit measurements on v0.16-era source include approximately 8.19 MB minified for D2, 1.87 MB for the PDF engine, 1.08 MB for the main index, and 753 KB for the editor. Lazy chunks still impose first-use latency and memory cost, while the main entry affects every launch.

## Proposed Solution

Establish measured bundle budgets, profile the main entry, and split optional subsystems at feature boundaries. Investigate lighter D2 delivery and avoid bundling unused syntax languages or exporter dependencies.

## Acceptance Criteria

- [x] CI records raw and gzip sizes for startup and major lazy chunks
- [x] A documented budget fails CI on meaningful regressions
- [x] Main startup chunk is reduced below the agreed budget
- [x] Optional AI, plugin marketplace, export, notebook, diagram, and editor code is absent from startup unless required
- [x] D2 delivery has a documented decision and materially smaller download or an explicitly accepted tradeoff
- [x] First-use loading states remain accessible and errors are recoverable
- [x] Before/after size and startup/first-use measurements are attached to the implementation PR

## Related

#177 provides Codecov bundle analysis. Prior lazy-loading work includes #99 and #163; this issue measures the current architecture rather than reopening those implementations.

## Implementation Tasks

- [x] Bundle size script (`scripts/check-bundle-size.mjs`) with `bundle-budgets.json` and unit tests
- [x] CI step on the ubuntu build leg that records raw/gzip sizes and fails over budget
- [x] Lazy-load the Sentry SDK behind the telemetry opt-in (local ErrorBoundary + report facade)
- [x] Lazy gemoji plugin (match the katex/highlight pattern)
- [x] Lazy AI chat panel
- [x] Lazy plugins modal and workspace settings modal
- [x] Cut static reach into `src/lib/export/*` at the feature boundary
- [x] D2 delivery investigation with decision recorded in `docs/bundle-budgets.md`
- [x] Final budgets measured and documented in `docs/bundle-budgets.md`
- [x] Before/after size table attached to the PR

hamidfzm:
## Summary

Adds a CI-enforced bundle budget gate and trims the startup chunk by 15% gzip. CI now records raw and gzip sizes for the startup assets and every major lazy chunk and fails the build when a documented budget is exceeded. The startup bundle drops from 381 kB to 323 kB gzip by making the Sentry SDK, the gemoji table, the AI chat panel, the plugin marketplace, workspace settings, and the export pipeline load on first use. The D2 delivery tradeoff is investigated and documented.

Closes #437

## Changes

- `scripts/check-bundle-size.mjs` + `bundle-budgets.json`: gzip budget gate over `dist/` (startup set parsed from `index.html`, lazy chunks matched by filename prefix, unbudgeted assets over 195 kB gzip rejected, loud failure on renamed/missing budgeted chunks or an empty startup set), with unit tests and a `pnpm size:check` script.
- `.github/workflows/ci-build.yml`: "Check bundle budgets" step on the `ubuntu-22.04` leg after the Tauri build, writing the size table to the step summary. No job renames, so the pinned required checks are untouched.
- `docs/bundle-budgets.md`: budgets rationale, update process, and the D2 delivery decision (upstream 0.1.33 ships a single 8 MB browser file with inlined WASM and no slimmer alternative; the chunk stays lazy behind `d2Render.ts` and its budget pins today's size).
- Sentry SDK is now loaded only after the production telemetry opt-in: `src/lib/telemetry.ts` imports `@sentry/react` on demand and exposes a `captureException` facade; a local `ErrorBoundary` (forwarding the React component stack) replaces `Sentry.ErrorBoundary` in `main.tsx`. Opt-out during the SDK download wins over the pending init, and re-enabling after an opt-out re-initializes the closed client.
- `remark-gemoji` loads lazily on the first `:shortcode:` sighting (`lazyGemoji` + `useGemojiPlugin`, mirroring the katex/highlight pattern); the pipeline takes a resolved `gemojiPlugin` and the site export sniffs content the same way.
- Lazy `lazy()` wrappers for the AI chat panel (first-open latch, panel stays mounted afterwards so composer draft and scroll survive), the plugins modal, and the workspace settings modal.
- `src/lib/export/*` is now reached only through dynamic imports (`useExport`, `usePrint`, plugin exporter runner); `d2Render.ts` imports DOMPurify on demand. Chunk-load failures land in the existing error logging instead of unhandled rejections.

## Risk classification

- [ ] Persistence / data loss
- [ ] Filesystem / IPC surface
- [ ] Secrets / credentials
- [ ] Network
- [ ] Accessibility
- [x] Bundle size / startup
- [ ] No risk areas touched

- **INV-3 (stale results never win), async ordering**: the telemetry opt-in is now async. An opt-out that lands while the SDK is still downloading wins over the pending init (`telemetry.test.ts` "does not init when the user opts out while the SDK is still loading"), and re-enabling after an opt-out re-initializes the closed client instead of silently dropping every later capture ("re-initializes when re-enabled after an opt-out closed the client"). The gemoji hook blocks stale completions: a plugin resolution that lands after the content stopped containing shortcodes cannot overwrite newer state (`useGemojiPlugin.test.ts` "goes back to null when the shortcode disappears").
- **INV-5 (all external input is untrusted)**: `sanitizeSvg` in `d2Render.ts` became async (DOMPurify imported on demand), but every render path still resolves through it before any SVG can reach `innerHTML`; if the DOMPurify import fails, the cached promise rejects and is evicted, so an unsanitized string is never observable. Covered by `d2Render.test.ts` (sanitizer wiring, `FORBID_TAGS: ["foreignObject"]`).
- **Bundle size / startup**: the point of the PR. The budget gate itself is the regression guard (`scripts/check-bundle-size.test.mjs` covers over-budget, missing-prefix, empty-startup, and unbudgeted-asset failures), and CI fails on any meaningful regression from here on.

## Measurements (before / after)

Production Vite build, measured by `scripts/check-bundle-size.mjs` (gzip level 9; raw in parentheses):

| Group | Before | After |
|---|---:|---:|
| **startup (entry JS + CSS)** | **381.0 kB (1228.4 kB)** | **323.0 kB (1057.5 kB)** |
| d2 (lazy) | 5849.0 kB (7997.1 kB) | unchanged |
| pdfEngine (lazy) | 807.5 kB (1825.0 kB) | unchanged |
| MarkdownEditor (lazy) | 263.3 kB (781.5 kB) | unchanged |
| mermaid (lazy) | 133.3 kB (576.5 kB) | unchanged |
| AI chat panel | in startup | 2.4 kB (6.4 kB), loads on first open |
| Plugins modal | in startup | 2.0 kB (6.8 kB), loads on first open |
| Workspace settings | in startup | 3.4 kB (13.3 kB), loads on first open |

Startup is 58 kB gzip (171 kB raw JS parse) lighter; that raw delta is the removed Sentry SDK, gemoji table, and AI/marketplace/export code every launch used to parse. First-use costs are the lazy-chunk sizes above; loading uses the same `Suspense fallback={null}` pattern as the existing lazy features, and a failed chunk load logs and aborts recoverably instead of rejecting unhandled. No wall-clock startup benchmark was run; the measured quantity is bundle size.

A full production `pnpm tauri build --no-bundle` on this branch passes with zero Cargo warnings, measures startup at 312.3 kB gzip against the 336.9 kB budget (about 7% headroom), and produces a 30.3 MiB release binary (`glyph.exe`).

## Testing

- [ ] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

Automated gates on the development machine (Windows): `pnpm typecheck`, `pnpm check`, `pnpm test` (3307 tests), `cargo clippy --all-targets -- -D warnings`, `pnpm test:coverage` (100% line coverage on every changed file), `pnpm build && pnpm size:check` (gate passes with the new budgets), and a production `pnpm tauri build --no-bundle` (success, zero Cargo warnings).

## Screenshots

Not applicable; the user-visible surface is unchanged (first-use loading follows the existing lazy-feature pattern).

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -62,6 +62,12 @@ jobs:
           CODECOV_TOKEN: ${{ matrix.bundle_analysis && secrets.CODECOV_TOKEN || '' }}
         with:
           args: ${{ github.event_name == 'pull_request' && matrix.args_pr || matrix.args_full }}
+      # Records raw/gzip sizes for startup and major lazy chunks and fails on
+      # budget regressions (bundle-budgets.json, docs/bundle-budgets.md). One
+      # deterministic leg is enough: every platform ships the same frontend.
+      - name: Check bundle budgets
+        if: matrix.platform == 'ubuntu-22.04'
+        run: node scripts/check-bundle-size.mjs
       # Boots the real webview under the production CSP and exercises the
       # grant-gated CLI export end to end; unit tests cannot enforce CSP.
       - name: Headless export smoke test
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,6 +99,7 @@ pnpm test                       # Run frontend tests (Vitest)
 pnpm test:watch                 # Run frontend tests in watch mode
 pnpm test:coverage              # Run frontend tests with coverage
 pnpm test:e2e                   # WebKit smoke tests (Playwright; run `pnpm exec playwright install webkit` once first)
+pnpm size:check                 # Bundle budget gate against dist/ (see docs/bundle-budgets.md)
 cd src-tauri && cargo check     # Rust type checking
 cd src-tauri && cargo clippy    # Rust linting
 cd src-tauri && cargo test      # Rust tests
--- a/bundle-budgets.json
+++ b/bundle-budgets.json
@@ -1,0 +1,16 @@
+{
+  "$comment": "Gzip byte budgets enforced by scripts/check-bundle-size.mjs in CI. Rationale and update process: docs/bundle-budgets.md",
+  "startup": 345000,
+  "chunks": {
+    "d2": { "prefix": "d2-", "budgetGzip": 6450000 },
+    "pdfEngine": { "prefix": "pdfEngine-", "budgetGzip": 910000 },
+    "MarkdownEditor": { "prefix": "MarkdownEditor-", "budgetGzip": 297000 },
+    "mermaid": { "prefix": "mermaid.core-", "budgetGzip": 150000 },
+    "cytoscape": { "prefix": "cytoscape.esm-", "budgetGzip": 156000 },
+    "docx": { "prefix": "docx-", "budgetGzip": 116000 },
+    "katex": { "prefix": "katex-", "budgetGzip": 93000 },
+    "epub": { "prefix": "epub-", "budgetGzip": 34000 },
+    "html2canvas": { "prefix": "html2canvas.esm-", "budgetGzip": 52000 }
+  },
+  "unbudgetedChunkGzipLimit": 200000
+}
--- a/docs/bundle-budgets.md
+++ b/docs/bundle-budgets.md
@@ -1,0 +1,76 @@
+# Bundle Budgets
+
+CI enforces gzip size budgets on the frontend bundle so size regressions fail a
+PR instead of shipping silently. The gate is `scripts/check-bundle-size.mjs`,
+run by the `Check bundle budgets` step of the `ubuntu-22.04` build leg after
+the frontend is built; budgets live in [`bundle-budgets.json`](../bundle-budgets.json).
+
+Run it locally against a fresh production build:
+
+```bash
+pnpm build && pnpm size:check
+```
+
+## What is measured
+
+- **startup**: every asset referenced from `dist/index.html` (the entry script,
+  stylesheets, and any modulepreloads). This is what every launch parses.
+- **Named lazy chunks**: each `chunks` entry in `bundle-budgets.json` matches
+  `dist/assets/` files by filename prefix and sums them. These load on first
+  use of their feature, so their cost is first-use latency plus installer size.
+- **Unbudgeted assets**: any other asset (chunk, stylesheet, font) above
+  `unbudgetedChunkGzipLimit` fails the check. A new heavyweight dependency must
+  get a stable chunk name (via `manualChunks` in `vite.config.ts` or its
+  natural file name) and an explicit budget before it can merge.
+
+Budgets are gzip bytes (gzip is what matters for install/transfer size; raw
+sizes are recorded in the CI step summary for parse-cost context). Each budget
+is the measured size plus roughly 10% headroom, so dependency bumps pass while
+a feature accidentally landing in the wrong chunk does not.
+
+## What keeps startup small
+
+The startup chunk is the markdown viewer shell: React, the remark/rehype
+pipeline, the Tauri API glue, i18n, and the app shell components. Everything
+optional loads on first use:
+
+- Editor (CodeMirror), notebook, canvas, graph view, and the settings modals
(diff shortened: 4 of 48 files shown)


hamidfzm:
The single partial line Codecov flags (useGemojiPlugin.ts line 19) is the unmount-cancellation guard inside the load effect. It mirrors the identical guard in useKatexPlugin, which carries the same partial on main, and it is not observable through the hook's public surface: the returned value is already gated on the content check, so a test around it would assert implementation rather than behavior. The codecov/patch status check passes.

hamidfzm:
Followed up on the patch-coverage partial in `useGemojiPlugin` rather than leaving it acknowledged.

f6c6b58 adds a test for a load that resolves after the content stopped needing it: the loader is stubbed with a deferred promise so the chunk can be held in flight while the document loses its shortcodes, and the test asserts nothing surfaces and that a later shortcode still resolves. Patch coverage on the file is now 100%.

One honest note from mutation-testing the change: the `if (!cancelled)` branch is redundant with the `contentHasShortcode ? plugin : null` return gate. Deleting either one alone keeps every test green, so no test can pin that branch specifically. It stays because `useKatexPlugin` and `useHighlightPlugin` have the identical shape, and making this hook the odd one out is worse than one inert line. The behavior the new test pins is the stale-load path itself, which the return gate provides and which test 3 also guards.

hamidfzm:
Pushed 7b6ef65: guards removed from all three hooks, and the interactive export window closed.

**Guards.** Dropped the `cancelled` flag from `useGemojiPlugin`, `useKatexPlugin`, and `useHighlightPlugin`. Mutation testing showed it redundant with each hook's return gate: deleting either one alone left every test green, and only the return gate has observable behavior (test 3 fails without it). One less inert line in three files, and no partial branches left.

**Interactive exports.** Fixing this properly meant fixing the gate, not just calling it. `waitForRenderIdle` decided "finished" from a quiet DOM plus filled diagram slots, but a document waiting on a lazy chunk mutates nothing, so quiet alone reported it done. Added a shared in-flight counter (`src/lib/markdown/pluginLoads.ts`) that the highlight, KaTeX, and gemoji loaders register with, and joined it to the settle condition. Interactive export, print, and plugin exporters now go through that gate like the CLI export already did.

Two regressions caught while wiring it up, both now pinned by tests:

- A canvas board is not a `.markdown-body`, so a board export waited out the full 15s deadline. The gate's root selector now includes `.glyph-canvas`, whose cards carry the same lazy content.
- A document closed while the native save dialog is open left nothing to wait for, turning an instant abort into a 15s stall. The interactive callers check for a rendered root before waiting.

Gates green: 3313 tests, typecheck, Biome, clippy, and 100% line/branch coverage across all 13 changed files. Budget gate still passes at 312.4 kB gzip startup against the 336.9 kB budget.

PR #650, opened 2026-08-18
feat(markdown): play video and audio, and carry them through exports

### LINKED ISSUE #600: Play video and audio in markdown documents
## Problem

Glyph renders local images from a workspace but has no video or audio playback.

- `<video src="clip.mp4" controls></video>` in a note renders a dead element. `MarkdownContent` has no `video` entry in its components map, so the relative `src` never goes through `resolveImageSrc` into an `asset:` URL and the webview resolves it against its own origin (404).
- `<audio>` is not in the sanitizer allowlist at all, so it is unwrapped before it reaches the DOM. Voice memos and recorded clips beside notes are ordinary content in a local-first workspace.
- The pieces are already half in place: `video` and `source` are allowlisted (from the raw-HTML work in #96 / #106), and the CSP already permits playback (`media-src 'self' asset: ... https: http:`). Only URL resolution is missing.

GitHub cannot do this at all: committed repository videos never render inline, and only drag-and-drop attachment URLs (`github.com/user-attachments/assets/...`) become a player. A local-first viewer can be strictly better here, at the cost of one small component.

## Proposed Solution

One `MediaComponent` mapped to both `video` and `audio` in `MarkdownContent`'s components map. Native browser controls, no custom player chrome: the built-in element already gives play/pause, scrubbing, volume, fullscreen, Picture-in-Picture, keyboard and media keys.

Rendering:

- `src/components/markdown/MediaComponent.tsx` resolves every URL attribute through the existing `resolveImageSrc` (`src`, `poster`, and each nested `<source src>`), so workspace-root confinement is inherited from `resolveWorkspacePath` rather than reimplemented.
- Elements render with `preload="none"`, so nothing is decoded until the user presses play.
- Asset-protocol scope needs no change: `grants.rs` already mirrors read grants into the asset scope with `allow_asset_dir`.

Sanitizer (`sanitizeSchema.ts`):

- Add `audio` to `tagNames`, with the same attribute allowlist as `video`.
- Add `preload` to the shared attribute list.
- Remove `autoplay` from the `video` allowlist and do not add it for `audio`. It is dead today because nothing renders these elements, but the moment a media component exists it lets a document start playback (and a remote fetch) on open, with no user gesture.

Exports, per target:

No export inlines media bytes as data URIs. `prepareContent.ts`'s `embedAsset` stays images-only: base64 inflates roughly 1.37x, `readAsDataURL` holds the whole file in memory, and a note with a screen recording would produce an unusable export. Exports also do not write sidecar files next to a single-file target: `grant_export_file` is exact-path write only, and `grants.rs` already asserts that a sibling of the export target is denied. Widening that grant is not worth a media copy, so no Rust changes are involved.

- HTML, DOCX, PDF: the media element becomes its poster image plus a link to the source, labelled with the file name. Without a poster, the link alone. The substitution happens once in `prepareContent.ts`, before the image pass, so the poster becomes an ordinary `` that the existing pass inlines and every downstream walker already understands: `docxInline.ts` emits `ImageRun` plus `ExternalHyperlink`, `htmlToPdf.ts` uses `imageNode`.
- EPUB: media under a configurable size limit is packaged into the zip with its media type in the OPF manifest and referenced relative. The bytes land inside the file the user picked, so no grant is involved. The limit is a setting (Off, 10, 50, 100 MB, default 10 MB); over the limit, or with embedding off, EPUB uses the same poster plus link fallback. Video has no EPUB3 core media type, so the fallback content also covers readers that refuse to play it.
- Site: `rewriteUrls.ts` already collects `img` sources into the copy pass through the generic `rewriteAssetRef`; extend that switch to media elements. This is the one target where media still plays, because the exporter copies the file into the output tree and the reference is rewritten relative.
- Remote sources keep their original URL in every target and are never downloaded at export time.

## Acceptance Criteria

- [x] `<video src="clip.mp4" controls></video>` in a workspace note plays inline with native controls
- [x] `<audio src="memo.mp3" controls></audio>` plays inline, that is `audio` survives sanitization
- [x] `poster` and nested `<source src>` resolve through the same path as `src`
- [x] A media URL resolving outside the workspace root renders nothing, with test coverage for `src`, `poster`, and `<source src>` separately
- [x] A remote `https:` media URL is passed through unchanged and plays
- [x] The sanitizer strips `autoplay`, so opening a document never starts playback
- [x] Rendered media elements carry `preload="none"`
- [x] Exports never inline media as data URIs, never write sidecar files, and never download a remote source
- [x] HTML, DOCX, and PDF exports show the poster image plus a link, or a link alone when there is no poster
- [x] EPUB export packages local media under the configured limit into the zip with a manifest entry, and falls back to poster plus link over the limit or when embedding is off
- [x] The EPUB media limit is a setting with an Off option, default 10 MB
- [x] Site export copies media files into the output and rewrites the reference to a relative path, and the media plays from the exported site
- [x] New user-visible strings exist in all five locales (`en`, `de`, `es`, `fa`, `zh`)
- [x] `samples/` demonstrates the syntax with a small committed media file
- [x] Gates pass: `pnpm typecheck && pnpm check && pnpm test`, plus `cargo clippy --all-targets -- -D warnings`

## Scope / Out of scope

In scope: `<video>` and `<audio>` rendering from raw HTML in markdown, local and remote sources, the sanitizer changes above, and every export target plus print.

Out of scope:

- `` image-syntax coercion. The HTML tag is explicit, and GitHub does not do this either. Can be revisited if requested.
- `<track>` subtitles: a third URL attribute and a third confinement gate for near-zero demand.
- Opening a media file as its own tab from the file tree. `IMAGE_EXTENSIONS` and `isSupportedFile` stay as they are.
- Custom player UI, playback-speed controls, playlists.
- YouTube and Vimeo iframe embeds. The CSP sets `frame-src 'none'`, and widening it is a separate decision.
- Inlining media bytes into an export as a data URI, and downloading remote media at export time.
- Sidecar files beside a single-file export, and the export write-grant widening that would need. HTML, DOCX, and PDF therefore never carry media; site export and an under-limit EPUB do.

## Implementation Tasks

- [x] Add `MediaComponent.tsx` and wire `video` and `audio` into `MarkdownContent`'s components map
- [x] Extend `resolveImageSrc` usage to `poster` and nested `<source>` children (helper in `src/lib/`, not inside the component file)
- [x] Sanitizer: add `audio` and `preload`, drop `autoplay`
- [x] Styles: sizing and dark-mode treatment for media blocks in the markdown body
- [x] Shared poster plus link fallback for the single-file export targets
- [x] Site export: media in the existing asset copy pass in `rewriteUrls.ts`
- [x] EPUB export: media packaged into the zip with manifest entries, under a configurable size limit
- [x] Setting for the EPUB media embed limit, plus its five-locale strings
- [x] DOCX and PDF exports: poster plus link fallback
- [x] Five-locale strings for the new setting. The fallback link is labelled with the file name, so it needs no translated copy.
- [x] Tests: resolution, escape attempts for all three URL attributes, `autoplay` stripping, and one test per export target
- [x] Docs: `samples/` entry demonstrating the syntax

## Affected areas

Markdown rendering

## Alternatives Considered

- Ship it as a plugin. Not viable: plugins are sandboxed with no access to `convertFileSrc`, so a plugin cannot produce `asset:` URLs for local files, which is the entire feature. The plugin markdown API also has no component-override hook for HTML tags, only remark, rehype, and fenced-block renderers.
- A custom player component. Adds keyboard, focus, and accessibility work to reimplement what the native element already does correctly.
- Convert to GIF, GitHub's recommended workaround. Loses audio and multiplies file size.

## Platform

All

hamidfzm:
## Summary

Video and audio in a note now play inside Glyph. `<video>` and `<audio>` render through components that resolve `src`, `poster`, and every `<source>` child with the same workspace-clamped resolver images already use, so a relative path beside the document just works and one that escapes the workspace renders nothing.

`video` and `source` were already in the sanitizer allowlist but nothing rendered them, so the tags were inert. `audio` was not allowlisted at all.

Exports keep the promise that a single file stays a single file: media is never inlined as a data URI, and no sidecar file is written next to the export target. Writing one would mean widening `grant_export_file` from exact-path write to a whole directory, which is not worth a media copy, so there are no Rust changes here. EPUB is the exception, because its zip can hold the bytes the user already picked a path for.

Closes #600

## Changes

- `MarkdownMedia` and `MarkdownMediaSource` resolve media references through `resolveAssetRef`, a new `{ src, path }` form of the existing image resolver that also reports the absolute path for the exporters. `resolveImageSrc` now delegates to it, so there is still one resolve-and-clamp path.
- Playback is forced to `preload="none"` and `controls`: decoding runs in the OS media stack, a much larger parsing surface than the image decoders, so an untrusted container stays unparsed until the user presses play.
- Sanitizer: `audio` added; `autoplay` and `preload` removed, and `poster` confined to http(s). `autoplay` was dead while nothing rendered these elements; with a media component it would let a document start playback, and a remote fetch, on open with no user gesture. `preload` is dead surface too, since both renderers set it themselves. `poster` had no protocol entry, and the website export is a bare unified pipeline with no url transform of its own, so a `file:` poster would have survived into a published page.
- `packageExportMedia` runs in `prepareContent` ahead of the link and image passes. It swaps each media element for a poster paragraph plus a paragraph naming the file (separate on purpose: the PDF walker only embeds an image that is a paragraph's sole child), so HTML, DOCX, and PDF need no walker changes. Media is named, never linked: no export carries the bytes, so a local file name would only resolve if the reader put the media back, and an exported file opened outside the app is better off carrying no href the document controls. The name is emphasised so it does not read as prose, since the class it carries is styled only in HTML and EPUB and both walkers drop it. A remote source is named by its full URL, which a reader can still copy. The exporter-only `data-media-path` is stripped from output, as is any orphan `<source>` whose asset URL would otherwise carry an absolute local path into the file.
- EPUB packages media under a configurable size limit into its zip with OPF manifest entries. New Setting in Print & PDF Export: Off, 10, 50, 100 MB, default 10 MB, honored by menu and CLI exports alike. The limit is the whole container's budget rather than a per-file ceiling, so a note of several near-limit recordings cannot be assembled into a zip that has to cross the IPC boundary as one array; entries are stored uncompressed (media is already compressed), a file referenced twice is packaged once, and hrefs are percent-encoded so a name with a space or `#` still resolves.
- Site export registers media and posters in the existing asset copy pass and forces the same playback attributes, since a static pipeline has no component to apply them. That is the one target where the player survives.
- Printing reuses the live DOM, so the export fallback never reached it and a player with no poster printed as an empty box. Each media element now renders a print-only span naming the file, hidden on screen and revealed by `print.css`; a player with a poster still prints its frame, and only the blank ones are dropped. The span carries `data-export-ignore`, which `prepareContent` already strips before building its own fallback, so no export names the media twice.
- `samples/` gains a Video and Audio section with a 24 KB clip and a 12 KB tone.

## Risk classification

- [x] Network
- [x] Migrations / persisted-format changes
- [ ] Persistence / data loss
- [ ] Asynchronous ordering / races
- [ ] Secrets / credentials
- [ ] Accessibility
- [ ] Bundle size / startup

**INV-5 (all external input is untrusted).** Media adds three new URL attributes a document can point anywhere: `src`, `poster`, and each `<source src>`. All three go through `resolveWorkspacePath`, the same clamp images use, rather than a new path. `MediaComponent.test.tsx` covers each attribute separately: a `src` escaping the root renders nothing, an escaping `poster` is dropped while the video survives, and an escaping `<source>` renders nothing. `sanitizeSchema.test.tsx` proves `autoplay` is stripped end to end and `preload="none"` is what reaches the DOM.

**INV-6 (the backend is the boundary).** No grant, command, or capability changed. Media loads through the asset protocol, whose runtime scope `grants.rs` already mirrors from the workspace read grant, and `media-src` in the CSP already allowed `asset:`. The sidecar-file option was dropped precisely because it would have needed a wider write grant.

**Network.** A remote `src` fetches when the user presses play, not on open, because `autoplay` is gone and `preload="none"`. Exports never download a remote source: `mediaAssets.test.ts` asserts `fetch` is not called for one, and that the exported document names it by its URL rather than linking it. A read that throws is a fallback, never an aborted export, so an unreachable network share cannot leave the user with no file and only a console error.

**Migrations.** `print.epubMediaLimit` is a new persisted setting. A settings file written by an older build has no value for it, so `epubMediaLimitBytes` returns 0 for anything non-numeric and the export links instead of embedding. Covered in `settings.test.ts`.

## Testing

- [ ] Tested on macOS
- [x] Tested on Windows
- [ ] Tested on Linux

Automated gates on Windows: `pnpm typecheck`, `pnpm check`, `pnpm test` (3436 tests, 383 files), and `cargo clippy --all-targets -- -D warnings`. New coverage: media resolution and the three escape paths, sanitizer behavior, the export fallback and packaging, the EPUB manifest, the site rewrite, and the settings control.

Site export playback is verified at the markup level (asset registered for copy, reference rewritten relative, controls present), not by loading the exported site in a browser.

## Screenshots

Not included: the change is a native media player rendered by the webview.


hamidfzm:
Second review pass on the diff, findings addressed in 9fe9a09 and the two commits before it.

Fixed:

- **Export could abort silently.** `readUnderLimit` called `fetch` unguarded, unlike its sibling `embedAsset`, which wraps it. An unreachable network share or a refused scheme rejected out through `prepareContent` into `useExport`'s catch, which only logs: the user picked a save path and no file was written. It now falls back to the link like any other unreadable media.
- **Absolute local path could leak into exports.** The media pass only walked `video, audio`, so an orphan `<source>` (the sanitizer unwraps `<picture>` and hoists its children) kept `src="asset://localhost/C%3A%2FUsers%2F..."` in exported HTML and EPUB. Every `<source>` still standing after the pass is now removed.
- **Empty player, and an empty anchor in exports.** The guard counted any element child, but a `<source>` whose src was refused still renders as a React element. `<video src="../../out.mp4"><source src="../../out.webm"></video>` left an empty framed player, and exported `<a href="">`, which in an HTML export points back at the document. The parent now reads the hast node to tell a resolved child from a refused one, and the fallback emits nothing when there is no file to name.
- **EPUB could be assembled beyond memory.** The size setting was a per-file ceiling with no aggregate, so three 90 MB clips at the 100 MB setting built a 270 MB zip, ran it through `Array.from` and JSON IPC, and deflated it on the main thread. The limit is now the whole container's budget, media is stored uncompressed, and a file referenced twice is packaged once.
- **File names broke EPUB hrefs.** `media/0-clip #2.mp4` truncates at the `#` and fails epubcheck; non-ASCII names produce IRIs. Hrefs are percent-encoded while the zip entry keeps the name.
- **`poster` had no protocol allowlist.** react-markdown's url transform covers the viewer, but the website export is a bare unified pipeline with none, so `poster="file:///C:/Users/.../private.png"` survived into a published page. Added `poster: ["http", "https"]` to the sanitizer, and dropped `preload` from the allowlist since both renderers set it themselves.
- Smaller: the `node` prop is stripped before spreading onto the DOM (it was stringifying to `node="[object Object]"`), the media pass moved above the anchor pass so a remote fallback link gets `target`/`rel`, the test file uses the shared `renderInWorkspace` helper and is renamed to match the module it imports, and the German and Spanish strings use "Größe" and "tamaño".

`mediaAssets.ts` is now at 100% lines and branches.

Two things the review raised that I did not change, for the record:

- **Print still shows the live player**, so a video without a poster prints as an empty box. Print reuses the live DOM by design, and giving it the export substitution would mean running that pass outside export. Worth a separate issue if it bothers anyone.
- **`connect-src` in the CSP does not list `asset:`**, only `http:`/`https:`, which covers Windows' `http://asset.localhost` form but not the `asset://localhost` form on macOS and Linux. That predates this PR (`embedAsset` fetches image assets the same way), and with the new guard the worst case is a graceful fallback to the link rather than a failure. Worth checking on macOS during review.

Issue #707, opened 2026-09-02
Website export leaves pages behind for content that no longer exists


hamidfzm:
### Description

The website export writes files but never removes the ones a previous export left behind, so an output directory that is exported into more than once accumulates pages for content that no longer exists.

Deleting `guide.md` and exporting again leaves `guide.html` in the output. It is gone from the navigation, because the nav is rebuilt from what is on disk now, but the file itself stays and still resolves at its own URL. A rename is a delete plus a create, so it leaves the old URL live alongside the new one.

This surfaced while verifying `glyph serve` (#703), where rebuilds happen constantly, but it is not a serve bug: `glyph <folder> --export site --out ./site` run twice into the same directory does the same thing. Serve only makes it obvious.

Two ways it bites:

- **Publishing.** Exporting into a directory that is deployed means deleted pages stay published, reachable by anyone holding or guessing the old link, and indexable. Removing a page from the workspace looks like it worked, because the nav no longer lists it.
- **Previewing.** `glyph serve` shows a site that does not match the folder, so a rename appears to have produced a duplicate page.

### Steps to Reproduce

1. Export a workspace as a website: `glyph ./docs --export site --out ./site`
2. Delete `docs/guide.md` (or rename it to `docs/handbook.md`)
3. Export again into the same directory: `glyph ./docs --export site --out ./site`
4. `./site/guide.html` is still there, with its old content

The same thing happens live under `glyph serve ./docs`: delete a file, and its page is still served at its URL after the rebuild.

### Expected Behavior

After an export, the output directory reflects only what the workspace currently contains. A page whose source is gone does not survive in the output, and a rename leaves only the new URL.

### Actual Behavior

Files from earlier exports stay in the output directory indefinitely. Only the navigation reflects the current workspace; the orphaned pages remain and keep resolving.

### Proposed Fix

`exportSite` (`src/lib/export/site/exportSite.ts`) already computes every path it writes, but returns only counts (`{ pages, assets }`). Have it report the site-relative paths it produced, then delete the files a previous export wrote that this one did not.

Prune against the previous build's own file list rather than against the whole directory. Wiping anything in the output directory that the current export did not produce would delete unrelated files a user keeps in their `--out` directory, including a `CNAME`, a `.nojekyll`, or anything else a static host needs. Deleting only what a previous export is known to have written keeps the blast radius to files Glyph created.

[...]

PR #708, opened 2026-09-02
feat(cli)!: add a serve subcommand, and make export one too

### LINKED ISSUE #703: Add a `glyph serve` command that watches a folder and serves the website export
### Problem

The website export is one-shot. `glyph <folder> --export site --out <dir>` renders the workspace, writes the files, and exits (`finish_cli_export` in `src-tauri/src/commands/export_runtime.rs`). To actually look at the result, or to show it to somebody, you need a second tool: a static file server pointed at the output directory, plus a manual re-export and a manual browser refresh after every edit.

That makes two common jobs awkward:

- **Previewing the export.** Checking that nav, themes, `.glyph/site.json` metadata, and rewritten links came out right means export, serve, refresh, repeat.
- **Sharing a folder of markdown.** Handing a colleague a URL for a docs folder currently means exporting, then uploading somewhere or running an unrelated web server by hand.

Glyph already owns every piece: the site renderer (`src/lib/export/site/exportSite.ts`), the CLI launch plan (`src-tauri/src/cli.rs`), and recursive directory watching (`src-tauri/src/watcher.rs`). Only the serving and the re-export loop are missing.

### Proposed Solution

A new CLI subcommand that exports a folder, keeps watching it, re-exports on change, and serves the result over HTTP with browser auto-reload:

```
glyph serve <dir> [--host <host>] [--port <port>] [--out <dir>]
```

**CLI parsing (`src-tauri/src/cli.rs`).** The CLI is flag-based with a single positional path and has no subcommands today. `serve` as `argv[1]` becomes a new parse branch producing a `CliLaunch::Serve { root, host, port, output }` variant alongside `Open` and `Export`. A folder literally named `serve` still opens normally via `glyph ./serve`, since only a bare first argument is treated as the subcommand.

**Dispatch (`src-tauri/src/setup.rs`).** The `Serve` arm mirrors the existing `ExportFormat::Site` arm: hidden window, `grant_workspace` on the root, `grant_export_dir` on the output. The difference is that the process stays alive instead of calling `finish_cli_export`.

**Rendering.** Serve mode reuses `exportSite` unchanged. The frontend runner (sibling to `src/hooks/useCliExport.ts`) exports once on mount, then re-exports whenever Rust emits a watch event, reporting success or failure back over IPC.

**Static server (new `src-tauri/src/serve.rs`).** `hyper` + `tower_http::services::ServeDir` over the output directory. ServeDir already handles MIME types, `index.html` resolution, and path-traversal containment, so none of that is hand-rolled. `hyper`, `tower-http`, `http-body-util`, and `tokio` are already in `src-tauri/Cargo.lock` transitively, so this adds no new dependency tree, only explicit `Cargo.toml` entries.

**Live reload.** A `GET /__glyph/reload` SSE endpoint backed by a `tokio::sync::broadcast` channel. A response layer appends a small script to `text/html` responses before `</body>`; the script reconnects and reloads on each event. After a successful re-export, Rust broadcasts one reload. A failed re-export keeps the last good site served, prints the error to stderr, and broadcasts nothing.

**Watching.** Recursive `notify` watch on the workspace root reusing `is_relevant_directory_change` from `src-tauri/src/watcher.rs`, with a short debounce so a burst of saves coalesces into one re-export.

**Output directory.** No `--out`: a temp directory, removed when the process exits. `--out <dir>`: that directory, kept after exit, exactly like `--export site --out`.

**Binding.** `--host` defaults to `127.0.0.1`. Any other host is an explicit opt-in and prints a warning to stderr that everyone on the network can read the workspace. `--port` defaults to `4173`; `--port 0` asks the OS for a free port, and the chosen port appears in the printed URL.

### Acceptance Criteria

- [x] `glyph serve <folder>` exports the folder, then prints the served URL on stdout (for example `Serving <folder> at http://127.0.0.1:4173`) and keeps running until interrupted.
- [x] `GET /` returns the exported index page; nested pages resolve at their site-relative paths; an unknown path returns 404.
- [x] Editing a markdown file under the served folder triggers a re-export, and a browser with the affected page open shows the new content without a manual refresh.
- [x] Adding, deleting, or renaming a markdown file re-exports and updates the navigation on the reloaded page.
- [x] A burst of rapid saves coalesces into a single re-export rather than one per event.
- [x] A re-export that fails leaves the previously served site intact, prints the error to stderr, and pushes no reload.
- [x] Without `--host`, the server binds `127.0.0.1` only and is unreachable from other machines.
- [x] `--host 0.0.0.0` binds all interfaces and prints a warning naming the exposure (the workspace becomes readable by anyone on the network).
- [x] `--port <n>` binds that port; a port already in use fails with a clear stderr message and a nonzero exit; `--port 0` binds an OS-assigned free port and prints it.
- [x] Without `--out`, output goes to a temp directory that no longer exists after the process exits. With `--out <dir>`, the exported files remain there.
- [x] `--out` pointing inside the served folder is rejected, matching the existing website-export rule that the destination must not be inside the workspace.
- [x] A traversal request (`/../../etc/passwd` and its percent-encoded and backslash forms) never serves a file outside the output directory.
- [x] `glyph serve` with a missing path, a file path, or a nonexistent directory prints a usage message on stderr and exits with code 2.
- [x] A folder named `serve` still opens interactively via `glyph ./serve`.
- [x] `glyph --help` documents the `serve` subcommand and its `--host`, `--port`, and `--out` flags.
- [x] The injected reload script appears only in served `text/html` responses, never in the files written to `--out`.

### Scope / Out of scope

**In scope**

- `glyph serve <dir>` subcommand with `--host`, `--port`, `--out`
- Static file server over the site export output, backed by `tower-http` ServeDir
- Recursive watch on the workspace root plus debounced full re-export
- SSE live reload with HTML script injection
- Temp output directory lifecycle
- `--help` text and README CLI documentation

**Out of scope**

- Any in-app UI for serve mode (menu item, status-bar indicator, stop control, settings). Worth a follow-up issue once the CLI shape settles.
- HTTPS/TLS, authentication, or access control of any kind
- Incremental re-export of only the changed page. A full re-export is the starting point; optimize later if it proves slow on large workspaces.
- Public tunneling, deployment, or hosting integrations
- Mobile (the CLI is desktop-only, matching the existing `tauri-plugin-cli` gating)

### Implementation Tasks

- [x] Add `CliLaunch::Serve` and the `serve` subcommand parse branch in `src-tauri/src/cli.rs`, with host/port/out parsing, validation, and unit tests
- [x] Dispatch the `Serve` arm in `src-tauri/src/setup.rs`: hidden window, workspace and export-dir grants, no process exit
- [x] Add `src-tauri/src/serve.rs`: hyper + `tower_http` ServeDir server, SSE reload endpoint, HTML script injection layer, broadcast channel
- [x] Wire a recursive debounced watch on the root reusing `is_relevant_directory_change`, plus the IPC commands the frontend runner needs
- [x] Add the frontend serve-mode runner beside `src/hooks/useCliExport.ts`, reusing `exportSite` and reporting outcomes back to Rust
- [x] Declare `hyper`, `tower-http`, `http-body-util`, `tokio`, and `tempfile` explicitly in `src-tauri/Cargo.toml` (all already transitive)
- [x] Temp output directory creation and cleanup on exit and on interrupt
- [x] Update `src-tauri/src/cli_help.rs` usage text and the README CLI section
- [x] Tests: CLI parse cases, path-traversal containment, script injection only on HTML, reload broadcast on success and silence on failure, debounce coalescing, frontend runner

### Affected areas

Rust backend, Build / CI

### Alternatives Considered

- **`glyph <dir> --serve`**, matching the existing `--export` flag shape. Rejected: serve mode is a long-running server, not a property of opening a document, and a subcommand reads correctly next to a directory argument.
- **Hand-rolled static server on `std::net::TcpListener`.** Rejected: MIME mapping and path-traversal containment are exactly the parts worth not writing, and `tower-http` ServeDir is already in the dependency graph.
- **Poll-based browser refresh (meta refresh or a polling script).** Rejected: it either reloads when nothing changed or lags behind edits. SSE is a few lines more and pushes exactly once per successful re-export.
- **Manual refresh only.** Rejected during scoping: without auto-reload the preview loop is barely better than exporting by hand.

### Platform

All

### LINKED ISSUE #710: Flip the Docker image and export action to the export subcommand
### Description

The CLI's export surface became a subcommand in #708 (spec: #703):

```bash
glyph <path> --export <format>          # before
glyph export <path> --format <format>   # after
```

Two consumers still called the old spelling on purpose, because they run the *released* binary rather than this repository's tip, and the release carrying the subcommand does not exist yet:

1. **The Docker image's default command** (`Dockerfile`) and the invocations in `.github/workflows/ci-docker.yml`.
2. **`glyph-md/export-website-action`**, updated in [glyph-md/export-website-action#2](https://github.com/glyph-md/export-website-action/pull/2) and deliberately not merged.

### Sequencing: no longer an issue for the image

This was originally filed as a timing problem in both directions. Flip before the release and the Docker check is red until it ships; flip after and the release's own image runs a command its bundled binary rejects.

That dilemma existed only because the Docker check installed whatever release happened to be newest. #715 removes the cause rather than picking a side: the Dockerfile's package download became a stage of its own, so CI substitutes the `.deb` the `ubuntu-22.04` build leg just produced, using buildx's `--build-context package=<dir>`. The release build is untouched and still downloads the published package.

The check now tests the pull request, so the flip lands with it and there is no window at all. The same fix means no future CLI change has to be sequenced against a release either.

Two related improvements came with it, since both were blocked by the same cause:

- The Docker workflow was path-filtered and therefore unrequirable, because a check that reports nothing on most pull requests blocks every merge when required. Its jobs now live in `ci-build.yml` beside Android and iOS, producing `Build / Docker` and `Build / Docker smoke`.
- The image build and the export smoke are separate jobs, so a broken image and a broken export are distinguishable failures.

**`glyph-md/export-website-action` is still genuinely release-coupled**, since it installs a published `.deb` at runtime and has no build leg to borrow from. It merges alongside the release.

### First publish to ghcr

No image had ever been published: `publish-docker` runs only on a `v*` tag, and no release had been cut since the Dockerfile landed in #705. v0.23.0 was the first push.

It was expected to need a manual visibility flip, on the basis that a container package created by `GITHUB_TOKEN` is private by default. **That did not happen.** The package came out publicly readable with no intervention: an anonymous `ghcr.io/v2/` token request succeeds, and the tag list and manifest both fetch without credentials. `docker/metadata-action` stamps `org.opencontainers.image.source`, which links the package to the repository, and the visibility followed the public repository.

Verified on the v0.23.0 push:

- tags `0.23.0`, `0.23`, `latest`
- an OCI image index carrying `linux/amd64` and `linux/arm64`
- both readable anonymously, so the `docker run ghcr.io/hamidfzm/glyph` in `README.md` works as documented

### Related pull requests

| PR | Role |
| --- | --- |
| #708 | Made `export` a subcommand, the breaking change this issue tracks (spec #703) |
| #705 | Added the Dockerfile and `ci-docker.yml` that still called the flag spelling |
| #715 | Flipped the image, its checks and the README, and moved the Docker jobs into `ci-build.yml` so they build this checkout's package. Covers the five checked tasks below |
| [glyph-md/export-website-action#2](https://github.com/glyph-md/export-website-action/pull/2) | The action's own flip, still open |

### Closing condition

v0.23.0 shipped on 2026-09-04, so the release gate is gone and every remaining task is actionable now.

The action shipped. Its smoke had failed on 2026-09-03 with `Glyph 0.22.1 has no glyph export subcommand`, which was its own precondition check firing correctly against the newest release at the time. A re-run against v0.23.0 passed both checks, exporting the fixture to `out/index.html` with wikilinks resolved, so that PR needed no change and merged as written.

**The release-notes callout is done, and so is the cause behind it.** v0.23.0's notes were already published, mentioning #708 only as its commit subject `feat(cli)!: add a serve subcommand, and make export one too`, so they were edited in place: the notes now open with an upgrade block giving the old and new spelling, the formats `--format` takes, and what breaks in scripts and in explicit `docker run` overrides.

Editing one release does not stop the next one repeating it. GitHub's generated notes group by label, so the `!` in a conventional-commit subject reaches nobody, and #708 landed under Features like any other entry. #736 adds a `breaking` label and a Breaking Changes category keyed off it, listed first, and applies the label to #708 retroactively. Since the generated bullet is only a PR title, CONTRIBUTING now also says a release carrying that section needs a hand-written upgrade note, pointing at v0.23.0's as the worked example.

The website docs task was not the flip it sounds like. `glyph-md/glyph-md.github.io` is a single marketing page (Hero, Features, Download, Footer) whose nav is Features, Download, GitHub. It had no docs section, and nothing on it mentioned the CLI, Docker, or the old flag spelling, so the task was new content rather than a rename.

That content is written and open as [glyph-md/glyph-md.github.io#8](https://github.com/glyph-md/glyph-md.github.io/pull/8): a Command line section after Download with four cards (export a document, export a website, serve, Docker), the full format list, and a link to the README's command-line section, translated into all five locales. It reuses the Download card and code-block styles, so the CSS change is three selector lists. That PR merged on 2026-09-08, so every task is done and this issue is closed.

### Tasks

- [x] Flip `Dockerfile`'s `CMD` to `["export", "/docs", "--format", "site", "--out", "/out"]` and drop the comment explaining the lag
- [x] Flip the two invocations in `.github/workflows/ci-docker.yml`
- [x] Flip the container example in `README.md` and drop the sentence explaining why it differs
- [x] Build the image against this checkout's package instead of the newest release, so the flip needs no release window
- [x] Add `Build / Docker` and `Build / Docker smoke` to ruleset 15172749
- [x] Merge [glyph-md/export-website-action#2](https://github.com/glyph-md/export-website-action/pull/2)
- [x] Cut an `export-website-action` tag matching the Glyph release, and decide whether its `v1` keeps moving or freezes at the last flag-era version. `v1` keeps moving (now at `268bf7c`); `v0.23.0` is the immutable tag naming the Glyph release it targets, and the [first release](https://github.com/glyph-md/export-website-action/releases/tag/v0.23.0) documents the break. The action also stopped installing `latest`: [#3](https://github.com/glyph-md/export-website-action/pull/3) pins `version` to `v0.23.0` by default, so the action and the binary it installs are a pair that was tested together
- [x] Update the website docs on `glyph-md/glyph-md.github.io` to the subcommand, and mention the document formats (`pdf`, `docx`, `epub`, `html`) alongside `site` (folded in from #625). Merged as [glyph-md/glyph-md.github.io#8](https://github.com/glyph-md/glyph-md.github.io/pull/8): a Command line section after Download with export, serve, and Docker cards, translated into all five locales, merged on 2026-09-08
- [x] Call the breaking CLI change out in the release notes. v0.23.0 now opens with an upgrade block showing the old and new spelling, the formats `--format` takes, and what breaks in scripts and `docker run` overrides. The generator itself was the deeper cause and is fixed in #736: GitHub groups notes by label, so `feat(cli)!:` reached nobody; there is now a `breaking` label and a Breaking Changes category listed first, and #708 carries the label
- [x] After the first release publishes the image, confirm it is anonymously pullable (it was already public; no flip needed)

### Affected areas

Build / CI

### Platform

All

hamidfzm:
## Summary

Adds a `serve` subcommand: render a folder as a website, keep serving it, and rebuild whenever the folder changes, with pages reloading themselves in the browser.

```bash
glyph serve ~/notes/                    # http://127.0.0.1:4173
glyph serve ~/notes/ --port 0           # let the OS pick a port
glyph serve ~/notes/ --host 0.0.0.0     # let the local network read it
glyph serve ~/notes/ --out ./site       # keep the build
```

Previewing the website export previously meant exporting, starting a separate static server, and refreshing by hand after every edit.

**This PR also renames the export surface**, because adding `serve` as a subcommand left the CLI with two shapes for one kind of job:

```bash
glyph <path> --export <format>          # before
glyph export <path> --format <format>   # after
```

`--export` is removed rather than deprecated, following the project's cleanup convention that a replaced path is deleted rather than left alongside its successor: a second accepted spelling would be the same inconsistency wearing a compatibility hat. Typing it now exits 2 with a message naming the replacement, since it is sitting in people's CI scripts.

**Breaking change, and it needs release notes.** Anyone running `--export` breaks on the next release. Updated here in the same delivery: help text, README, the Docker image's default command, and the CI smoke tests. The satellite action is [glyph-md/export-website-action#2](https://github.com/glyph-md/export-website-action/pull/2), which must merge **after** the release carrying this, not before, or it breaks every caller on `version: latest`.

Closes #703

## Changes

- **CLI** (`src-tauri/src/cli.rs`): `export` and `serve` subcommands sharing one parser. A subcommand counts only as a bare first argument, so a folder named after one still opens as `glyph ./serve`. One path scanner strips flag values before looking for the positional, so `glyph export --out build docs` cannot mistake `build` for the path, and one predicate decides whether a launch does its own work rather than forwarding to a running window. A flag with no subcommand is now a usage error instead of a launch that quietly ignores it. The plugin's declared `export`/`out` args are gone as dead config: everything but the positional path is read from argv.
- **Server** (`src-tauri/src/serve/`): `tower-http`'s `ServeDir` behind hyper, an SSE reload route, and script injection per response. Every crate was already in the tree through Tauri's own HTTP stack; enabling the `fs` feature adds a few small ones (`mime_guess`, `http-range-header`).
- **Watch** (`src-tauri/src/serve/watch.rs`): recursive `notify` watch, 300ms debounce so a burst of saves is one rebuild.
- **Split of work**: the site pipeline only exists in the renderer, so Rust owns the socket, the watch and process lifecycle, while `useCliServe` owns the export and reports back over `serve_ready` / `serve_failed`.
- **`useExportReadiness`**: the settings-and-plugins gate, extracted from `useCliExport` so both headless paths share one copy rather than two.
- Also excludes both subcommands from single-instance forwarding. Without it, `glyph serve` while the app was open exited 0 having served nothing while the running app opened the folder in a tab.

## Risk classification

- [x] Network
- [x] Build / CI

**INV-3 (a stale result never overwrites a newer one).** Changes arriving mid-build coalesce into exactly one more build: `building` / `queued` are closure-scoped in the single effect, every path funnels through one `build`, and `serve_ready` is awaited before the loop re-enters, so an older completion cannot land after a newer one. Covered by "collapses changes that arrive mid-build into one extra build" in `useCliServe.test.ts`, which holds the first export open, fires three changes, and asserts two builds rather than four.

**INV-5 / INV-6 (untrusted input, the backend is the boundary).** Every containment check is in Rust, not the renderer:

- Traversal: `/../`, `/%2e%2e/`, `/%2e%2e%2f`, `/..%5c` and `/notes/../../` are all refused, asserted against a real file planted outside the served directory (`serve/mod.rs`, `traversal_never_escapes_the_served_directory`).
- **DNS rebinding**: loopback binding alone does not keep a workspace private, because a page on any site can re-point its own hostname at 127.0.0.1 and have the browser treat the result as same-origin. Requests are answered only when `Host` names this server. Tested both ways, and verified live: `-H "Host: evil.com"` gets 403 while `localhost` gets 200.
- Hidden paths are never served, so an `--out` pointing at a working directory does not publish `.env` or `.git/config`.

**INV-7 (a partial result is never presented as complete).** A failed rebuild reports to stderr and pushes no reload, so browsers keep the last version they loaded. This is deliberately *not* claimed to be the previous site in full: the export writes pages in place, so a failure part way through leaves new and old pages mixed. That is #707, filed separately, and the comments, help text and README say what actually happens rather than overstating it.

**Network.** The app's first inbound listener. `docs/security/threat-model.md` gains an entry covering the loopback default, the `--host` opt-in and its warning, the `Host` check, and the fact that the whole `--out` directory is published (which is why an `--out` containing, or contained by, the served folder is refused). What is served is unauthenticated by design: anyone who can reach the port reads the exported site.

**Filesystem.** The temp directory comes from `tempfile` (random name, owner-only) rather than a name derived from the process id under a world-writable temp directory, where a local user could pre-create it or redirect it with a symlink and so choose the HTML the victim's browser loads from localhost.

## Testing

Gates green: 3767 frontend tests, 570 Rust tests, typecheck, Biome, `cargo fmt`, and clippy with `-D warnings`. Codecov reports all modified lines covered.

Verified against the built binary, not only unit tests:

- Ready line prints the folder and URL, process stays up. `GET /` serves the index with the reload script injected; nested pages resolve; unknown paths 404.
- Editing a file pushed `data: reload` on the SSE stream, with the new content served immediately after. Adding a file serves it and adds it to the nav; deleting removes it from the nav.
- `--port 0` bound an OS-assigned port and printed it; a busy port exits 1 with the OS message; usage errors exit 2.
- `--out` files survive the process; the reload script is absent from the files on disk (`grep -c` returns 0), so a site written with `--out` stays a plain static site.
- After the review fixes: `Host: evil.com` 403, `localhost` 200, `/.git/config` 404, HEAD reports the true file size rather than an injected one, live reload still working.
- After the rename, against a rebuilt binary: `glyph export <folder> --format site --out <dir>` and `glyph export <file.md> --format html --out <path>` both exit 0 and write their output; `--help` documents both subcommands; an unknown format and the removed `--export` flag each exit 2, the latter naming its replacement.

A code review pass ran before this PR and found ten warnings; eight are fixed in the second commit, and the other two were documentation overstating behavior, now corrected. The worst was a filter that judged hidden paths absolutely, so serving a folder under a hidden one (`~/.config/notes`, a dotfiles repo) built once and then silently never rebuilt again.

- [ ] Tested on macOS
- [x] Tested on Windows
- [ ] Tested on Linux

## Screenshots

Not applicable: the process renders headlessly and shows no window.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -76,18 +76,18 @@ jobs:
         env:
           WEBKIT_DISABLE_COMPOSITING_MODE: "1"
         run: |
-          xvfb-run --auto-servernum ./src-tauri/target/release/glyph "$PWD/samples" --export site --out /tmp/csp-smoke
+          xvfb-run --auto-servernum ./src-tauri/target/release/glyph export "$PWD/samples" --format site --out /tmp/csp-smoke
           test -f /tmp/csp-smoke/index.html
           test "$(find /tmp/csp-smoke -maxdepth 1 -name '*.html' | wc -l)" -ge 3
           # A document export renders the live DOM, so it exercises the
           # render-readiness gate the site export never touches.
-          xvfb-run --auto-servernum ./src-tauri/target/release/glyph "$PWD/samples/Index.md" --export html --out /tmp/doc-smoke.html
+          xvfb-run --auto-servernum ./src-tauri/target/release/glyph export "$PWD/samples/Index.md" --format html --out /tmp/doc-smoke.html
           grep -q "<h1" /tmp/doc-smoke.html
-          xvfb-run --auto-servernum ./src-tauri/target/release/glyph "$PWD/samples/Index.md" --export pdf --out /tmp/doc-smoke.pdf
+          xvfb-run --auto-servernum ./src-tauri/target/release/glyph export "$PWD/samples/Index.md" --format pdf --out /tmp/doc-smoke.pdf
           test "$(stat -c%s /tmp/doc-smoke.pdf)" -gt 5000
           # A wrong input kind is a usage error, not a silent empty export.
-          if xvfb-run --auto-servernum ./src-tauri/target/release/glyph "$PWD/samples" --export pdf --out /tmp/nope.pdf; then
-            echo "expected a usage error for a folder given to --export pdf" >&2
+          if xvfb-run --auto-servernum ./src-tauri/target/release/glyph export "$PWD/samples" --format pdf --out /tmp/nope.pdf; then
+            echo "expected a usage error for a folder given to --format pdf" >&2
             exit 1
           fi
       # Drives the built binary over WebDriver (launch, render, edit, second
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -100,7 +100,7 @@ jobs:
           docker run --rm --user "$(id -u):$(id -g)" \
             -v "$PWD/samples:/docs:ro" \
             -v "$RUNNER_TEMP/doc:/out" \
-            glyph:ci export /docs/Index.md --format html --out /out/index.html
+            glyph:ci /docs/Index.md --export html --out /out/index.html
           test -s "$RUNNER_TEMP/doc/index.html"
 
       - name: A rejected export fails the step
@@ -110,7 +110,7 @@ jobs:
           if docker run --rm --user "$(id -u):$(id -g)" \
             -v "$PWD/samples:/docs:ro" \
             -v "$RUNNER_TEMP/doc:/out" \
-            glyph:ci export /docs --format pdf --out /out/nope.pdf; then
+            glyph:ci /docs --export pdf --out /out/nope.pdf; then
             echo "::error::exporting a folder as pdf should have failed"
             exit 1
           fi
--- a/.github/workflows/ci-docker.yml
+++ b/.github/workflows/ci-docker.yml
@@ -100,7 +100,7 @@ jobs:
           docker run --rm --user "$(id -u):$(id -g)" \
             -v "$PWD/samples:/docs:ro" \
             -v "$RUNNER_TEMP/doc:/out" \
-            glyph:ci /docs/Index.md --export html --out /out/index.html
+            glyph:ci export /docs/Index.md --format html --out /out/index.html
           test -s "$RUNNER_TEMP/doc/index.html"
 
       - name: A rejected export fails the step
@@ -110,7 +110,7 @@ jobs:
           if docker run --rm --user "$(id -u):$(id -g)" \
             -v "$PWD/samples:/docs:ro" \
             -v "$RUNNER_TEMP/doc:/out" \
-            glyph:ci /docs --export pdf --out /out/nope.pdf; then
+            glyph:ci export /docs --format pdf --out /out/nope.pdf; then
             echo "::error::exporting a folder as pdf should have failed"
             exit 1
           fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -75,5 +75,12 @@ WORKDIR /docs
 # The container's arguments are Glyph's arguments, so every documented flag
 # works here without a second CLI to learn. The default renders the folder
 # mounted at /docs into /out.
+#
+# This spelling tracks the *released* CLI, not this checkout's: the image
+# installs a published .deb, so the default command may only use syntax that
+# release already has. `glyph export <path> --format site` replaces it, and
+# flipping it here has to wait for the release that carries the subcommand.
+# The Docker CI job is what enforces that, by building against the latest
+# release and running this command.
 ENTRYPOINT ["/usr/local/bin/glyph-entrypoint"]
-CMD ["export", "/docs", "--format", "site", "--out", "/out"]
+CMD ["/docs", "--export", "site", "--out", "/out"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,4 +76,4 @@ WORKDIR /docs
 # works here without a second CLI to learn. The default renders the folder
 # mounted at /docs into /out.
 ENTRYPOINT ["/usr/local/bin/glyph-entrypoint"]
-CMD ["/docs", "--export", "site", "--out", "/out"]
+CMD ["export", "/docs", "--format", "site", "--out", "/out"]
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ glyph serve ~/notes/ --host 0.0.0.0       # let the local network read it
 glyph serve ~/notes/ --out ./site         # keep the build instead of a temp dir
 ```
 
-`serve` renders the folder, prints the URL, and keeps running: every change to the folder rebuilds the site, and pages open in a browser reload themselves. It binds `127.0.0.1` unless `--host` says otherwise, so nothing leaves the machine by default. Without `--out` the site lives in a temporary directory that is removed on exit.
+`serve` renders the folder, prints the URL, and keeps running: every change to the folder rebuilds the site, and pages open in a browser reload themselves. It binds `127.0.0.1` unless `--host` says otherwise, so nothing leaves the machine by default, and it answers only to its own hostnames so a web page cannot reach it through your browser. Everything in the output directory is served, so point `--out` at a directory of its own; without one, the site lives in a temporary directory removed when you interrupt the command. Like the exports below, it renders through a webview, so a Linux server without a display needs `xvfb-run`.
 
 `--export` accepts `pdf`, `docx`, `epub`, `html`, and `site`. Without `--out` a document export writes beside its input with the format's extension; `site` always needs one. Exports run without showing a window, print the path they wrote to stdout, and exit nonzero with a message on stderr if they fail, so they can drive CI publishing (on Linux runners, wrap the command in `xvfb-run`). A document export includes the table of contents when Settings > Print has it enabled.
 
--- a/README.md
+++ b/README.md
@@ -179,6 +179,16 @@ glyph notes.md --export docx --out ~/out.docx
 glyph ~/notes/ --export site --out ./site
 ```
 
+```bash
+# serve a workspace as a website, rebuilding as it changes
+glyph serve ~/notes/
+glyph serve ~/notes/ --port 8080          # pick the port; 0 asks the OS
+glyph serve ~/notes/ --host 0.0.0.0       # let the local network read it
+glyph serve ~/notes/ --out ./site         # keep the build instead of a temp dir
+```
+
+`serve` renders the folder, prints the URL, and keeps running: every change to the folder rebuilds the site, and pages open in a browser reload themselves. It binds `127.0.0.1` unless `--host` says otherwise, so nothing leaves the machine by default. Without `--out` the site lives in a temporary directory that is removed on exit.
+
 `--export` accepts `pdf`, `docx`, `epub`, `html`, and `site`. Without `--out` a document export writes beside its input with the format's extension; `site` always needs one. Exports run without showing a window, print the path they wrote to stdout, and exit nonzero with a message on stderr if they fail, so they can drive CI publishing (on Linux runners, wrap the command in `xvfb-run`). A document export includes the table of contents when Settings > Print has it enabled.
 
 The command is provided by the Homebrew cask (macOS), Chocolatey or Scoop (Windows), and the deb package or Homebrew formula (Linux). The macOS `.dmg` and Windows MSI install the app only; use a package manager for the terminal command.
--- a/README.md
+++ b/README.md
@@ -208,10 +208,10 @@ docker run --rm --user "$(id -u):$(id -g)" \
 ```bash
 docker run --rm --user "$(id -u):$(id -g)" \
   -v "$PWD/docs:/docs:ro" -v "$PWD/out:/out" \
-  ghcr.io/hamidfzm/glyph export /docs/README.md --format pdf --out /out/readme.pdf
+  ghcr.io/hamidfzm/glyph /docs/README.md --export pdf --out /out/readme.pdf
 ```
 
-The image ships the same `.deb` the apt repository serves, plus the WebKitGTK stack and the X server a headless export needs, so no `xvfb-run` wrapper is required. It runs as uid 1000; passing `--user` keeps exported files owned by you when your uid differs. Images are published for `linux/amd64` and `linux/arm64` on every release, tagged `X.Y.Z`, `X.Y`, and `latest`.
+The image ships the same `.deb` the apt repository serves, plus the WebKitGTK stack and the X server a headless export needs, so no `xvfb-run` wrapper is required. Its arguments are the released CLI's, which is why the example above still spells the export as a flag: the image tracks the published release, not this repository's tip. It runs as uid 1000; passing `--user` keeps exported files owned by you when your uid differs. Images are published for `linux/amd64` and `linux/arm64` on every release, tagged `X.Y.Z`, `X.Y`, and `latest`.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -172,11 +172,11 @@ glyph ~/notes/       # open a folder as a workspace
 glyph --help         # usage, flags, and export formats
 
 # export a document (headless; exits when done)
-glyph README.md --export pdf              # writes README.pdf beside it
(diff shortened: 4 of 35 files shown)


hamidfzm:
Patch coverage started at 90.59% with 84 lines missing, almost all of it the parts only reachable over a live socket and a live directory. Rather than adding them to the ignore list, the last two commits test them for real: the server tests bind an ephemeral port and speak HTTP/1.1 down a socket, and the watcher test writes to a real directory and asserts that a burst of saves arrives as one rebuild.

That took it to 98.45%. Coverage by file: `serve/mod.rs` 85% to 98%, `serve/watch.rs` 84% to 99%, `commands/serve.rs` 90% to 99%.

The 16 lines left are IO failures that need a broken socket to reach, and one loop that by construction never returns:

- `serve/mod.rs` 101-108: `set_nonblocking` and `TcpListener::from_std` failing on a socket that was just bound successfully.
- `serve/mod.rs` 119-122: the `accept` error arm, including its back-off. Reaching it means exhausting file descriptors mid-test.
- `serve/mod.rs` 137: the closing brace of `run`'s accept loop, which only runs until the process ends.
- `serve/mod.rs` 218: a response body that fails part way through being collected, after `ServeDir` has already read the file from disk.
- `commands/serve.rs` 61 and `cli.rs` 315: two guards that return a safe answer when a mutex is poisoned or a path cannot be canonicalized.
- `serve/watch.rs` 91: the `continue` for an event the filter rejects. The filter itself is covered directly, including the hidden-path cases.

Flagging these rather than leaving the report unexplained. If any of them is worth a fault-injection test instead, say so and I will add it.

PR #722, opened 2026-09-04
fix(editor): refuse Save As onto a path another tab already holds

### LINKED ISSUE #721: Refuse Save As onto a path another tab already holds
## Problem

`saveVirtualAs` in `src/hooks/useDocumentSave.ts` writes the chosen Save As path onto the virtual tab without checking whether another tab already holds that path. `openFile` dedupes by path, so Save As is the one way two tabs end up on the same file.

Steps to reproduce:

1. Open `notes.md`.
2. New document, type something.
3. Save As, pick that same `notes.md`.

Two tabs now show `notes.md`, with independent buffers and independent dirty state. Consequences:

- Closing either tab calls `unwatch_file` unconditionally (`closeTabs`, `src/hooks/useTabs.ts:308`), and the watcher registry is not refcounted (noted at `src/hooks/useOpenDocument.ts:146`), so the surviving tab silently loses auto-reload.
- With Auto Save on, both tabs write to the same path. Last writer wins, and neither tab sees the other's write, because each write marks a self-save and the watcher echo is suppressed.
- The session snapshot records the path twice, and restore routes through `openFile`, which dedupes, so one of the two tabs quietly disappears on the next launch.
- The `file-changed` reload had to grow a per-tab dirty guard so a reload could not overwrite the second tab's unsaved buffer (#720). That guard contains the damage; it does not stop the state from arising.

The Save As write itself is not the loss here. It overwrites a file the user explicitly chose, which is what Save As means. The problem is the duplicate tab it leaves behind.

## Proposed Solution

Refuse the collision before writing anything. After `pickSave` returns a target, check whether another file tab already holds it; if so, surface a workspace notice naming the file and return `false` without calling `write_file`.

Collapsing the two tabs instead (write, then fold the virtual tab into the existing one) was considered and is worse: when the already-open tab is dirty, the collapse has to discard its unsaved buffer, which is exactly the silent discard INV-1 forbids. Refusing before the write cannot lose anything, since nothing has happened yet.

## Acceptance Criteria

- [ ] Save As onto a path another file tab already holds does not write to disk.
- [ ] It surfaces a workspace notice naming the file, and the virtual tab stays open and unsaved so the user can pick a different name.
- [ ] The already-open tab keeps its content, its dirty state, and its file watch.
- [ ] Save As onto a free path is unchanged.
- [ ] Tests cover the refusal and the untouched existing tab.

## Scope

In scope: the collision check in `saveVirtualAs`, its notice string, and tests.

Out of scope: refcounting the backend watcher registry; whether `repointOpenFiles` can produce the same collision through rename or move (not investigated).

## Implementation Tasks

- [ ] Add the collision check and notice key to `src/hooks/useDocumentSave.ts` and `src/locales/*/workspace.json`
- [ ] Tests in `src/hooks/useTabs.virtual.test.tsx`
- [ ] Confirm the per-tab guard added in #720 still has a reachable case, and note it in the PR either way

## Affected areas

layout / app-shell

## Platform

All

hamidfzm:
## Summary

`saveVirtualAs` wrote the Save As target onto the virtual tab without checking whether another tab already held that path. `openFile` activates the existing tab rather than opening a second one, so Save As was the only way two tabs could end up on one file, and that state is bad in several directions at once: closing either tab calls `unwatch_file` for the shared path and the registry is not refcounted, so the survivor silently loses auto-reload; with Auto Save on both tabs write over each other and neither sees the other's write; and the session snapshot records the path twice, so restore (which dedupes) quietly drops one tab on the next launch.

This refuses the collision after the dialog returns and before the write, surfaces a notice naming the file, and leaves the buffer virtual so the user can pick a different name.

Refusing before the write is the only non-destructive option. Folding the two tabs together after writing sounds friendlier, but when the already-open tab is dirty the fold has to throw away its unsaved buffer, which is the silent discard INV-1 forbids. Nothing has happened yet at the point of the check, so nothing can be lost.

Closes #721

## Changes

- `src/hooks/useDocumentSave.ts`: collision check between `pickSave` and `write_file`, returning `false` with a persistent workspace notice.
- `src/locales/{en,de,es,fa,zh}/workspace.json`: `notice.saveTargetOpen` in all five locales.
- `src/hooks/useTabs.virtual.test.tsx`: the refusal, with the already-open tab deliberately dirty so the test also pins that its buffer survives.

## Risk classification

- [x] Persistence / data loss
- [ ] Asynchronous ordering / races
- [ ] Untrusted rendering (Markdown, plugins, links)
- [ ] Secrets / credentials
- [ ] Network
- [ ] Accessibility
- [ ] Bundle size / startup
- [ ] No risk areas touched

**INV-1: user edits are never silently discarded.** The refusal is what protects the already-open tab's unsaved buffer, and it protects the virtual buffer too by leaving it virtual and dirty rather than half-adopting a path. `useTabs.virtual.test.tsx` "saveDocument refuses a Save As onto a path another tab already holds" asserts both buffers survive and that `write_file` was never called; it fails without the guard (`expected true to be false`, the save reporting success).

**Filesystem surface.** The change only removes a write that used to happen. No new command, no new path handling, and `basename` for the notice is the same helper the neighbouring `saveFailed` notice uses.

### Note on #720

[#720](https://github.com/hamidfzm/glyph/pull/720) fixes a stale edit buffer on the same reload path, and it is now stacked on this branch: its refresh is only safe while one tab per path holds, which is what this PR establishes.

The producers are covered. `openFile` activates the existing tab instead of opening a second one, this PR makes Save As refuse a collision, and `repointOpenFiles` cannot reach it either, because `rename_path` and `move_path` both route their target through `unique_path` (`src-tauri/src/commands/create.rs`), which bumps to "name 1" rather than landing on an existing file.

**Merge this one first.** #720 targets this branch, and GitHub retargets it to `main` once this lands.

## Testing

`pnpm typecheck && pnpm check && pnpm test` (3777 tests, 407 files, all passing) plus `cargo clippy --all-targets -- -D warnings` via the pre-commit hook. The new guard was verified load-bearing by removing it and watching the test fail. `localeParity.test.ts` covers the five-locale requirement.

- [ ] Tested on macOS
- [x] Tested on Windows
- [ ] Tested on Linux

## Screenshots

n/a


hamidfzm:
Folded into #720, which now carries both commits and closes #721. Closing this one so there is a single PR to review.

Issue #301, opened 2026-06-14
Implement an MCP server that exposes Glyph's vault index


hamidfzm:
## Problem

AI agents already read markdown files. What they cannot do is understand a vault the way Glyph does. Concrete failures of `cat` and `grep` on a Glyph or Obsidian vault:

- `[[Graph View]]` is a stem, not a path. Grep misses `[[graph view]]`, `[[Graph View|the graph]]`, `[[Graph View#Layout]]`, and `![[Graph View]]`, and cannot say which of two `Graph View.md` files wins (the one in the same directory does).
- "What links here" needs every link in the vault resolved against this file. Grepping the stem returns prose mentions and fenced code, and misses case and alias variants.
- `#alpha` in a code fence, `issue #42`, and `mid#word` are not tags; `tags: work, ideas` is two tags; `a/b` nests under `a`. The vault index encodes these rules, grep does not.
- Frontmatter is YAML parsed with the same scalar rules the renderer uses. Regex over `---` blocks drifts.
- A `.canvas` file is a JSON blob to an agent and a graph of nodes and edges to Glyph.
- Which vault is open, which note is active, what tabs are up: not on disk in any file an agent would think to read.
- Rendering to PDF, DOCX, EPUB, HTML, or a website is Glyph's pipeline, not something an agent can reproduce from files.

Glyph has all of this, but only the person looking at the window can use it. MCP clients get nothing.

**Where the incumbents are.** Obsidian shipped an official CLI in 1.12 that exposes backlinks, outgoing links, unresolved links, orphans, properties, tags, Bases, tasks, and daily notes, and community MCP servers wrap it. It is a remote control for a running app: if Obsidian is closed, the first command launches it. Joplin ships an MCP server inside the app on the Web Clipper port, read tools on by default and write tools off by default with a per-tool setting; it too needs the app running. Typora, MarkText, and Zettlr have nothing. Glyph can match the link and metadata surface and beat both on one axis: the index works headless, in one binary, with no port and no token.

## Proposed Solution

Ship an MCP server as a subcommand of the existing binary, and make every tool pass one test: **an agent with filesystem tools cannot get this answer, or gets a wrong one.** Tools that fail the test do not ship.

### CLI surface

```
glyph mcp [--vault <root>]...    Serve the vault index to an MCP client over stdio
```

[...]

PR #729, opened 2026-09-08
feat(vault): read the workspace index from Rust and delete the TypeScript one

### LINKED ISSUE #223: Move the workspace link index to Rust as the single source for the UI and the MCP server
## Problem

Several existing and planned features need the same data: every wikilink, every backlink, every alias, every tag, every heading, every block ID across the workspace. Today that data is split across the process boundary in the worst place:

- Rust does the raw scanning: `scan_wikilinks` (unresolved refs with line and snippet), `scan_metadata` (raw frontmatter block plus inline tags), `search_workspace`, and `walk.rs` with its caps and `ScanStatus`.
- The webview holds everything that makes the scan useful: the wikilink resolver (`wikilinkResolver.ts`), the graph model (`graph.ts`), the metadata index and tag rules (`metadata.ts`), the backlinks filter (`backlinks.ts`), heading slicing (`headingSection.ts`), the palette's property filters (`metadataQuery.ts`), and canvas parsing (`canvas/parse.ts`).

Twenty modules consume that TypeScript layer (graph view, backlinks section, files panel, tags pane, embeds, hover preview, tab context, wikilink completion, workspace scan). Two costs follow:

1. **Drift by construction.** Constants are mirrored by hand across the boundary (`metadata.ts` notes that it "mirrors SCAN_MAX_TAG_CHARS in metadata.rs"), and every new feature that touches links ships its own subtly different parser. This is the duplication the original version of this issue was filed against.
2. **Nothing without a webview can use the index.** The MCP server (#301) runs as `glyph mcp`, a headless Rust process that never boots a webview. As the index stands, the only way for it to resolve a link or list backlinks would be a second implementation in Rust, which is exactly the drift above.

The original plan was a TypeScript module, with a Rust port "if profiling later shows the indexer dominating". #301 changes the order: the port is the prerequisite, not the optimisation.

## Proposed Solution

One `vault` module in `src-tauri/src/`, the single source of truth for the workspace, consumed in-process by the MCP server and over one command surface by the UI.

**Owns, per workspace root:**

- The file set, via the existing walker, with the existing caps and `ScanStatus` truncation reporting.
- Per note: title, parsed frontmatter, tags (frontmatter and inline, with the rules `metadata.rs` and `metadata.ts` encode today: fences, `issue #42`, `mid#word`, scalar list splitting, `a/b` nesting, caps), headings with level and line (fence-aware), outgoing links (raw target, resolved path, heading fragment, alias text, embed flag), and later block IDs (#218) and tasks (#712).
- The resolver: case-insensitive stem match, `.md` stripped, same-directory tie-break, `#heading` and `|alias` split, `![[embed]]`. Keyed on `aliases:` too, so #217 becomes a lookup rather than a feature.
- Derived views: backlinks per note, the graph (nodes, edges, degree, orphan flag), unresolved links, dead-ends.
- The property query grammar from `metadataQuery.ts` (`tag:foo`, `status:draft`, quoted values), ported so the palette and #713 share it.
- Canvas parsing into nodes and edges.

**Updates:** built on workspace open, kept current by the existing `notify` watcher on create, edit, delete, and rename, without a full rescan.

**Consumers:** the UI reads the index through Tauri commands (a snapshot per root plus targeted queries: resolve, backlinks, neighbours, tags, query). The webview keeps only rendering-time concerns such as building wikilink nodes from resolution results. The TypeScript resolver, graph builder, metadata index, and backlinks filter are deleted, not kept as fallbacks. `scan_wikilinks` and `scan_metadata` are replaced, not wrapped.

**Parity:** a shared fixture vault used by the Rust tests, with the existing TypeScript tests either moved to Rust or rewritten to assert against command output, so the port cannot silently change what a link resolves to.

**Frontmatter:** Rust has no YAML dependency today. The parser must match the renderer's `js-yaml` FAILSAFE semantics (every scalar stays a string, so `2026-04-15` and `true` render verbatim). Crate choice (maintained fork of serde_yaml, or yaml-rust2) is a planning decision on maintenance status and binary size.

## Acceptance Criteria

- [x] The Rust resolver matches the TypeScript resolver on the shared fixture vault: case, same-directory tie-break, `#heading`, `|alias`, `![[embed]]`, `.md` suffix, ambiguous stems
- [x] Tags reproduce today's rules (fences, `issue #42`, `mid#word`, scalar list split, nesting, caps), and the mirrored constants are gone from TypeScript
- [x] Frontmatter parsed in Rust yields the fields and values the renderer shows, with dates and booleans kept as strings
- [x] Backlinks section, tags pane, graph view, hover preview, embeds, wikilink autocomplete, and palette property filters read from the Rust index, and their existing tests pass against it
- [x] Creating, editing, deleting, or renaming a file updates the index incrementally and consumers refresh without a manual action
- [x] `ScanStatus` truncation survives the move and is surfaced as today (#436, #479)
- [x] The index is usable in-process with no webview and no Tauri app handle; `glyph mcp` is the first such caller
- [x] No TypeScript module keeps a parallel implementation of resolution, tag rules, graph building, or the query grammar
- [x] Indexing a 1k-note fixture is no slower than today's scan on the same machine, measured in the PR

## Scope / Out of scope

**In scope:** the `vault` module, resolver, metadata and tags, headings, links, backlinks, graph, unresolved and dead-end detection, canvas parsing, the query grammar port, the UI consumer switch, watcher-driven updates, the shared fixture vault.

**Out of scope:**

- The MCP transport and tools (#301), which consume this
- `aliases:` resolution (#217), a lookup once this lands
- Block references (#218) and tasks (#712), which add columns to this index later
- The properties table view (#713) and link rewriting on rename (#711), which build on it
- The typed vault event bus (#227)
- Full-text ranking and semantic search (#372)

## Implementation Tasks

Split across two pull requests (see the implementation plan comment). The first leaves the app on the existing scan commands; the second switches consumers and deletes the TypeScript index.

**PR A, the Rust core (`Part of #223`)**

- [x] `vault` module skeleton: file set through the shared walker, caps, `ScanStatus`, `.canvas` included
- [x] `yaml-rust2` dependency and the FAILSAFE frontmatter parser
- [x] Per-note extraction: title, frontmatter fields, tags, headings, outgoing links
- [x] Tag rules ported whole: fences, `issue #42`, `mid#word`, scalar split, `a/b` nesting, both caps
- [x] Resolver with stem, nested-suffix, and `aliases:` keys, and the same-directory tie-break
- [x] Derived views: backlinks, graph with degree and orphan, unresolved, dead-ends
- [x] Query grammar port
- [x] Canvas parsing into index nodes and edges
- [x] Command surface, each wrapper checking the `GrantRegistry` first
- [x] Shared fixture vault and the parity tests
- [x] Performance measurement on a one-thousand-note fixture, old path versus new

**PR B, the switch (`Closes #223`)**

- [x] `useWorkspaceIndex` reads `vault_snapshot`; `WorkspaceIndexStatus` collapses to files plus vault
- [x] Backlinks section, tags pane, files panel, graph view onto the snapshot and targeted queries
- [x] `useWikilinkResolutions` and the `wikilinkNodes` lookup change; hover preview and embeds follow
- [x] Palette onto `vault_query`
- [x] Watcher-driven incremental updates
- [x] Delete the five TypeScript modules, their tests, and the two Rust scan commands
- [x] Confirm the truncation banner still fires, and record the performance numbers in the PR

## Affected areas

Rust backend, layout / app-shell, markdown rendering.

## Alternatives Considered

- Keep the TypeScript index and give the MCP server its own Rust resolver: two resolvers, two tag parsers, guaranteed drift.
- Run the MCP server inside the webview so it can reach the TypeScript index: needs the app running, a local port, and an auth surface, and still leaves the mirrored constants in place.

## Platform

All

## Additional Context

Blocks #301, #711, #712, #713, #217, #219. The original text of this issue proposed a TypeScript module with a Rust port deferred to profiling (see #186 for the same call on the file ranker); the MCP server made the port the first step.

hamidfzm:
## Summary

Second half of #223. #728 built the Rust `vault` index and its command surface but left the app running on the old scans; this switches every consumer onto it and deletes the TypeScript index that used to derive the same answers.

**+2075 / -2635 across 100 files.** Five TypeScript modules and their tests are gone, along with the two Rust scan commands they were built on.

## Changes

- **One snapshot per root.** `useWorkspaceIndex` calls `vault_refresh` when a workspace opens, `vault_snapshot` after a watcher change, and `vault_forget` when it closes. It still runs `list_markdown_files` alongside: that is every openable document (markdown, notebooks, canvases, D2), which is what the palette, autocomplete, the embed gate and the auto-open probe mean by "a file here", while the index covers the notes it parses (markdown and canvases). `WorkspaceIndexStatus` is those two walks.
- **Targeted queries for what the snapshot omits.** `useBacklinks`, `useTaggedPaths` and `useVaultQuery` replace `filterBacklinks`, `pathsWithTag`, and `parseMetadataQuery` + `matchesFilters`. Backlinks stay a query because a workspace has far more links than files and each carries a snippet of up to 200 characters.
- **Resolution left the remark pass.** `buildLinkNode` and `buildEmbedNode` are map lookups against a `resolutions` map threaded through `RemarkPipelineOptions`. `MarkdownContent` fills it from `useWikilinkResolutions`; `exportSite` awaits `resolveTargets`, because `renderPage.ts` renders markdown entirely outside React. The pipeline stays synchronous.
- **Deleted:** `wikilinkResolver.ts`, `graph.ts`, `metadata.ts`, `backlinks.ts`, `metadataQuery.ts` and their five test files; `scan_wikilinks` and `scan_metadata` with their command modules; the mirrored tag constants (`metadata.ts` used to note that it "mirrors SCAN_MAX_TAG_CHARS in metadata.rs").
- **The build reads files in parallel.** The walk returns paths and the files are read and indexed across the available cores; see Performance.
- **Two helpers survived rather than being reimplemented:** `stemOf` became `pathStem` over `stem` in `paths.ts`, and `isNestedTarget` moved into `wikilinkNodes.ts`, which owns wikilink syntax.

## Performance

Same 1k-note fixture, same machine, warm runs.

| | old | new |
| --- | --- | --- |
| Rust: build the index | `scan_wikilinks` + `scan_metadata` 133 / 141 ms | `Vault::build` 24 / 28 ms |
| TypeScript: derive from it | metadata 2.3 / 3.3 ms, tags + fields 0.6 / 1.0 ms, **graph 79 / 81 ms** | none |
| Cross the boundary | two scan payloads | one 914 KB snapshot, 1.7 / 1.8 ms to serialize |
| **Total** | **≈ 215 to 226 ms** | **≈ 26 to 30 ms** |

Two things account for it. The old `buildWorkspaceGraph` re-resolved every wikilink in JavaScript, on the UI thread; that work now happens once in Rust while the parsed notes are already in hand. And within the build, reading the files was nearly all of it: walking 0.8 ms, parsing 4 ms, resolving and building the graph 1.4 ms, reading 55 to 63 ms on one thread. Each file is an open, a read and a close, and none depends on another until resolution, so the read now runs across the available cores and the build went from 66 to 72 ms to 24 to 28 ms.

The backlink dedup also stopped being quadratic: it scanned every existing row for the target on each link, which a note with thousands of inbound links would feel. Links arrive in source then line order, so only the last row can be a duplicate.

Reproduce with `cargo test --release --lib vault::tests::index_1k_notes_timing -- --ignored --nocapture`, which prints the phase breakdown too.

## Risk classification

- [x] Destructive lifecycle (close, unmount, workspace switch, app exit)

**INV-3 (stale results never win).** Every index read is asynchronous now, so several paths could land on the wrong workspace, note or document.

- `scanWorkspace` and `refreshIndexes` drop an answer whose workspace was replaced, and a monotonic walk counter means the newest walk wins rather than the last one to return. A change arriving *during* the opening walk is remembered and the walk repeated: the copy that walk built predates the write, and an incremental update only re-reads paths it is told about. Covered in `useWorkspaceIndex.test.ts` and `useTabs.workspace.test.tsx`.
- `useBacklinks`, `useTaggedPaths` and `useWikilinkResolutions` stamp each answer with the subject it answers. The resolution stamp matters most: `SplitView` is not keyed by tab, so without it the next document's first render would use the previous one's paths and a click or `![[embed]]` could open the wrong note, past the membership gate, because the path is a real workspace file. Covered in each hook's test file.

**INV-4 (owners flush before they die).** `clearIndexes` sends `vault_forget`, on close and on a workspace switch. The tree drops in the same batch, or a surviving consumer (a loose tab, an open palette) re-asks about the root just forgotten and the backend rebuilds and re-caches it. `useTabs.workspace.test.tsx` asserts the forget on both paths and that nothing asks about the outgoing root afterwards.

**INV-7 (partial results explicit).** The truncation banner still fires, keyed so a dismissed banner re-shows only when the truncation changes. A document the index has not answered for yet is *pending*, not broken: links render plain and embeds show the loading placeholder, rather than every link flashing `wikilink--broken` and every embed "not found" on each open and tab switch.

**IPC surface.** Two commands removed, none added; `docs/security/threat-model.md` drops their row.

## Testing

`pnpm typecheck && pnpm check && pnpm test` (412 files, 3856 tests), `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (700 tests) all pass. Patch coverage is clean: every stale-answer arm in the index hooks has a test that lets the current answer land first and then settles the stale ones.

Roughly 250 existing cases moved onto the index rather than being deleted, including the whole `wikilink.test.ts` suite (now keyed on resolution maps), the graph suites (a shared `graphOf` fixture), and the site-export suites.

- [ ] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Notes for review

- `graphNeighbors.ts` is TypeScript deriving something from graph data, which the acceptance criteria restrict. It turns the index's edge list into an adjacency map for hover dimming and decides nothing: which links exist, which resolve, and which collapse into one edge are all settled in Rust. `tagTree.ts` is the same shape over `tagCounts`. Flagging both rather than having them found.
- `[[analysis]]` naming `analysis.ipynb` no longer resolves. The old TypeScript resolver was handed the openable-document list, so notebooks and D2 files resolved incidentally; the Rust resolver only knows notes. That behaviour was undocumented and untested, and making attachments resolvable means giving them ids inside the index, which is a change to the index rather than to its consumers. Left out deliberately.
- `vault_neighbors` and `vault_canvas` still have no frontend caller. They exist for the MCP server in #301, which is the point of the port.
- #741 unified the extension lists behind `src-tauri/extensions.json` and touches `is_supported_file`, the predicate separating the two walks. They share a cap but count different files against it, so a workspace heavy in notebooks can truncate one and not the other.

Closes #223


hamidfzm:
A review pass over the diff turned up two regressions and a set of stale-result holes. All fixed in 05a3e40; the gates and the 1k-note timing were re-run after.

### The workspace has two file sets, and this had collapsed them into one

`snapshot.files` is the note index: markdown and canvases, because those are the files the index parses. The list it replaced came from `list_markdown_files`, which is every openable document, so it also carries notebooks and D2 sources.

Everything reading `workspaceFiles` means the second set: the palette's Files rows, wikilink autocomplete, the AI system prompt's file list, the embed and hover-preview membership gate, and `files.includes(remembered)` in the auto-open probe, where a notebook silently opened a different note instead.

`workspaceFiles` reads `list_markdown_files` again, and `WorkspaceIndexStatus` is `{ files, vault }`. That is what the issue's task list actually said ("`WorkspaceIndexStatus` collapses to files plus vault"); I over-collapsed it.

One narrowing stands and is worth a look: `[[analysis]]` naming `analysis.ipynb` used to resolve, because the old TypeScript resolver was handed the openable list. The Rust resolver only knows notes. The old behaviour was undocumented and untested (the deleted resolver's own header said "a flat list of workspace **markdown** file paths"), and making attachments resolvable means giving them ids in the note index, which is a change to the index rather than to its consumers. Left out deliberately; say if it should be in scope here instead of a follow-up.

### `glyph serve` resolved links against the index as of process start

`vault_resolve` builds the index on first use and caches it per root. The serve watcher only emits `serve://changed`; unlike the app watcher it never calls `apply_changes`, and nothing on the serve path refreshes or forgets. So after the first build, a note created since was a permanently broken link and a renamed one kept resolving to a page that was no longer generated. Before this PR each rebuild re-derived resolution from a fresh `list_markdown_files`, so serve was always current.

`exportSite` now refreshes the index once before the page loop, which fixes the plain export path too.

### Stale answers could win in four places

- **Two directory changes in a row** applied whichever walk *returned* last, not the newest one started. A monotonic walk counter decides now.
- **A watcher snapshot built during the opening walk** was then replaced by that walk's older result in the backend cache, and the file it saw was never re-read, so it stayed invisible for the session. The refresh is skipped while an open is in flight for that root.
- **Resolutions carried over between documents.** `MarkdownViewer` is keyed by tab, but `SplitView` is not, so the first render of the next document used the previous one's map: a click or an `![[embed]]` could open the wrong note, and the membership gate would not catch it because the path is a real workspace file. The map is stamped with the document it answers, so a different one gets nothing rather than the last one's paths. Same for closing the workspace under a loose tab, which used to keep resolving into the closed workspace forever.
- **Backlinks and tagged paths** showed the previous note's or tag's rows under the new heading for one round trip.

### Three more

- **The outgoing workspace's index was never released** on a folder switch, only on close. A session that opened A then B then C held three full indexes (notes, tags, resolver, graph) for the rest of the process.
- **Resolutions never re-asked when the index changed**, so creating a link's target left it broken until the document's text changed, and renaming one left the anchor pointing at the old path. `snapshot` is now a dependency, and the ask is keyed on the target list rather than the document body, so typing prose between two links no longer re-asks.
- **`vault_query` returned every path in the workspace** when the query carried no filters, on every keystroke, and the palette threw it away. It also ran while the palette was closed, re-firing on every watcher refresh for the life of the window. Rust returns paths only when there are filters, and the hook is gated on the palette being open.

### Tests

`useWikilinkResolutions` had no test file; it now has eight cases including the wrong-document guard and the re-ask on re-index. `useWorkspaceIndex` gets its own file for the two ordering races and the release paths, `useTabs.workspace.test.tsx` asserts `vault_forget` on both close and replace, and `useBacklinks` covers the switch-to-another-note case.

3750 frontend tests and 680 Rust tests pass; typecheck, Biome, `cargo fmt --check` and strict clippy are clean. Patch coverage still has no missing lines.

### Two notes the review raised that I did not change

- `stem(".gitignore")` returns the empty string, which is a live bug in the rename box (`FileTreeEntry.tsx`). Pre-existing and unrelated to this PR, so it wants its own issue rather than a drive-by fix here.
- `vault_neighbors` and `vault_canvas` still have no frontend caller. They exist for the MCP server in #301, which is the point of the port; `useCanvasDocument` keeps parsing boards in TypeScript because the canvas editor round-trips the file it edits, which the index deliberately does not own.

hamidfzm:
Rebased onto `main` (through #736). The conflict was in `exportSite.test.ts`, where #734 and #735 added `switch` cases next to the ones this branch adds; both sides kept.

A second review pass found five more things, three of them mine from the last round.

### The guard I added for a mid-open change dropped it instead of deferring it

Last round I stopped a watcher-driven read from landing during the opening walk, because `vault_refresh` replaces the backend's copy wholesale and the read would have been thrown away. What I actually shipped was a permanent drop.

On a workspace large enough for the walk to outlast the 300ms debounce, a file written meanwhile (a sync client, a `git pull`) is absent from the index for the rest of the session: the walk's copy predates the write, and an incremental update only re-reads paths it is told about. Its link renders broken, the palette and autocomplete lack the file, the graph lacks the node. Before this PR the same event re-ran the scans from disk, so it is a regression. The event is now remembered and the walk repeated, and the test that pinned the drop now pins the re-walk.

### Switching folders released the outgoing index and immediately rebuilt it

`clearIndexes(previous.root)` sends `vault_forget(A)` and empties the snapshot, but the tree still names A at that point, and every query hook takes the snapshot as its re-ask trigger. So any consumer that survives the switch (a loose tab that became active, a document with `[[links]]`, an open palette) re-asked about A, and `with_vault` rebuilt and re-cached it. That reopened the INV-4 leak the last round closed, and put a full walk of the old workspace on the main thread mid-switch. The tree now drops in the same batch as the index, which is what `closeWorkspace` already did.

### Every document painted with all its links broken for a frame

Resolution is a round trip now, and an unanswered document was indistinguishable from one whose links do not resolve. So each open, tab switch, hover preview and nested embed flashed `wikilink--broken` and "Embedded note not found", then swapped. That reads as breakage rather than as loading, and for embeds it is a layout shift. Pending is now its own state: links render plain, embeds show the loading placeholder, and only an actual `null` from the index marks broken.

### Two smaller ones

`stem()` and `pathStem()` were not equivalent after all. #730 fixed `stem` for dotfiles, which is why I collapsed them, but `stem` needs a character after the dot and `pathStem` did not, so a file named `Note.` completed to `[[Note.]]`, which the Rust resolver keys as `note` and cannot resolve. `stem` now treats a trailing dot as an empty extension, matching `stem_of` in the resolver, and `pathStem` is one line over it.

`rankCommands` also collapses inner whitespace, which the deleted TypeScript parser did and the loading fallback does not, so `weekly  review` matched nothing while the first answer was in flight.

### Suppressions

Three of the four `useExhaustiveDependencies` suppressions are gone. An empty index has no backlinks, no tags and nothing to resolve, so the hooks read the snapshot to skip the round trip, and reading it is what makes the dependency honest rather than suppressed. The fourth stays with a specific reason: the query grammar recognises `field:` only for fields the index knows, so a re-index can change how the same query parses, and the hook never inspects that.

Also: no `vault_resolve` at all when the wikilink extension is off.

3774 frontend tests and 690 Rust tests pass; typecheck, Biome, `cargo fmt --check` and strict clippy clean. New tests cover the deferred re-walk, the clear-under-flight drop, the no-re-ask-after-forget rule, the pending link and embed, and the subject-switch case for tagged paths.

hamidfzm:
Rebased onto `main` again, through #738. This one had real conflicts rather than adjacent edits: #740 and #738 rewrote the same graph files this PR did, from the other direction.

They changed how the graph view *behaves* (camera and layout persisted per workspace root, click to focus, double press to open, zoom commands moved into their own hook). This PR changes where the graph *comes from* (the Rust index instead of `buildWorkspaceGraph` over `workspaceFiles` + `wikilinkRefs`). Both kept:

- `GraphView` takes `graph` from the snapshot and keeps the `persistKey` wiring plus the remount-by-root key in `TabContent`.
- Hover dimming reads `neighborIndex(graph.edges)` where it used to read `graph.neighbors` from the deleted builder, and it feeds the focus highlight (`highlightId`) rather than the raw hover id.
- `registerZoomTarget` is gone from `GraphView`, since #738 moved that into `useGraphZoomCommands`.
- The test fixtures move to `graphOf`: `TRIO_FILES` becomes `TRIO_GRAPH`, one unlinked note so focusing "a" still leaves something outside its neighbourhood to dim.

#741 landed in between and unified the extension lists behind `src-tauri/extensions.json`. It touches `is_supported_file`, which is the predicate separating the openable-document walk from the note index, so that split is worth a second look from whoever reviews: the two walks share a cap but count different files against it.

3842 frontend tests and 698 Rust tests pass; typecheck, Biome, `cargo fmt --check` and strict clippy clean.

PR #765, opened 2026-09-15
feat(vault): rewrite links when a note or folder is renamed or moved

### LINKED ISSUE #711: Rewrite links when a note or folder is renamed or moved
## Problem

Renaming or moving a note breaks every link that points to it. The file tree renames inline and drag-and-drop moves files and folders (#279), but nothing rewrites the `[[wikilinks]]`, `![[embeds]]`, or relative markdown links in other notes, so the backlinks panel goes empty and the graph loses its edges. Obsidian rewrites links on rename, its CLI does the same for `move` and `rename`, and every Obsidian MCP server exposes those as tools with link updates. Glyph leaves the user to grep and fix by hand.

The breakage differs by link kind, which is why a naive find-and-replace is wrong:

- A rename changes the filename stem, so `[[old-name]]`, `[[old-name|shown text]]`, `[[old-name#Heading]]`, and `![[old-name]]` all stop resolving.
- A move keeps the stem, so wikilinks keep working unless another file shares the stem, in which case the same-directory tie-break can silently pick a different target. Relative markdown links such as `[text](../old-name.md)` break on every move.
- `[[old-name]]` inside a code fence is not a link and must not be touched.

## Proposed Solution

Rewrite links from the workspace index, not from text search.

- On rename or move (file tree inline rename, drag-and-drop move, folder rename or move covering every note inside), ask the Rust workspace index (#223) for every inbound link that **resolves** to the affected file, so a prose mention or a same-stem file elsewhere is never rewritten.
- Rewrite each occurrence in place, preserving the parts the user wrote: alias text after `|`, `#heading` fragments, and the embed `!`. Wikilinks get the new stem; relative markdown links get a recomputed relative path from the linking file.
- Show a confirmation listing the affected files and link counts before writing. Writes go through the existing document save path so the self-save tracker and the file watcher treat them as Glyph's own edits.
- Open tabs that point at rewritten files reload their content the way external edits do today.
- After the rewrite, the index updates so backlinks and the graph reflect the new name immediately.

## Acceptance Criteria

- [x] Renaming a note rewrites `[[stem]]`, `[[stem|alias]]`, `[[stem#heading]]`, and `![[stem]]` in every note that resolved to it, keeping alias, heading, and embed marker intact
- [x] Moving a note rewrites relative markdown links to it and leaves wikilinks untouched unless the move changes what they resolve to
- [x] Moving a note recomputes its own relative markdown links (to notes and attachments) so they still resolve
- [x] Moving or renaming a folder rewrites links to every note inside it
- [x] Canvas file cards pointing at a renamed or moved note are rewritten, keeping `subpath` and the canvas file's formatting
- [x] A `[[stem]]` inside a code fence or inline code is never modified
- [x] A note whose stem matches but that did not resolve to the moved file (same stem in another folder) is never modified
- [x] The user sees the list of files about to change and can cancel; cancelling leaves every file byte-identical
- [x] An affected file open with unsaved changes is saved before the rewrite, and the confirmation says so
- [x] Rewritten files are saved through the normal save path, so no "file changed on disk" prompt appears for them
- [x] Backlinks panel and graph view show the new name without a manual refresh
- [x] Tests cover each link form, the same-stem trap, the fence case, folder moves, and a rewrite that fails midway (files already written stay written, the error names the file that failed)

## Scope / Out of scope

**In scope:** rename and move from the file tree and drag-and-drop, single notes and folders, wikilink and relative markdown link forms above (inbound, plus the moved note's own relative links), and canvas file cards.

**Out of scope:**

- Detecting renames made outside Glyph (the watcher sees a delete and a create, and cannot know they are one file).
- Undo beyond cancelling at the confirmation step; workspace sync history covers rollback.
- The agent-facing `rename_note` and `move_note` tools. They belong to the MCP write-tools issue that follows #301 and will call the same rewrite.
- Aliases in frontmatter (#217) and block references (#218): the rewrite handles whatever link forms the index resolves at the time it lands.

## Implementation Tasks

- [x] Scanner records link target byte ranges, skips inline code, and collects relative markdown link destinations
- [x] `apply_changes` handles folder paths so folder renames and moves update the index
- [x] Rewrite planner for wikilinks and embeds (shortest resolving target, alias, heading, and embed marker kept)
- [x] Rewrite planner for relative markdown links, inbound and outgoing of the moved note
- [x] Rewrite planner for canvas file cards
- [x] `rename_path` / `move_path` dry run and apply (rename, then rewrite), with an explicit partial-failure outcome
- [x] Shared tab reload helper
- [x] Confirmation dialog listing files and counts, marking unsaved files, with cancel
- [x] Wire rename and move: confirm, save affected dirty tabs, apply, reload open tabs, refresh the index, failure notice
- [x] Tests for every link form, the same-stem trap, fences and inline code, folder moves, canvas cards, and partial failure
- [x] Locale strings for the confirmation in all five locales

## Affected areas

Rust backend, layout / app-shell, markdown rendering (link forms).

## Alternatives Considered

- Text search and replace on the old stem: rewrites prose mentions, code fences, and same-stem files in other folders. This is the bug the index exists to prevent.
- Leaving links alone and relying on stem resolution: works for moves within the vault until two files share a stem, and never works for renames or relative markdown links.

## Platform

All

## Additional Context

Blocked by #223, which moves the link index to Rust. The MCP server in #301 exposes the same index read-only; write tools including rename and move come in a later issue and reuse this rewrite.

hamidfzm:
## Summary

Renaming or moving a note or folder from the file tree (inline rename, drag-and-drop, or "Move to...") now rewrites the links that point at it, after the user confirms. Wikilinks, embeds, relative markdown links (including the moved note's own relative links to notes and attachments) and canvas file cards are covered. Only links whose target would actually change are rewritten, so a same-named note in another folder, and links inside code, are never touched.

Closes #711

## Changes

- **Rewrite planner (`src-tauri/src/vault/relink.rs`)**: each link is resolved twice, against the workspace before the move and after it.
  - Wikilinks and embeds use a second `Resolver` built from the moved paths and get the shortest target that resolves to the new path. Alias, `#heading`, embed marker and a typed `.md` are kept.
  - Relative markdown links and canvas card `file` values resolve against the linking file's or board's folder, exactly as the renderer reads them (`normalizeRelativePath`), and get a recomputed relative path. `#fragment`, `<...>` destinations and a leading `./` are kept. On Windows and macOS a link spelled in another case still follows the move.
  - A link is left as written when no form of it can reach the new path: a note moved to the root while a same-named note sits beside the link, or a new name containing `#`, `|` or brackets that a link cannot spell.
  - Edits are byte-range splices, so a file stays byte-identical outside the rewritten targets (BOM and CRLF included).
- **Apply (`relocate`)**: syncs the index, plans from disk, renames first (the collision-safe final name only exists after the rename, and a failed rename writes nothing), then writes each rewrite through `ensure_writable`. A file that changed on disk since the plan read it is not overwritten. The first failed write stops the run and is reported in `failed`, leaving earlier files written, then the index is updated.
- **Commands**: `rename_path` and `move_path` take `dryRun`, return `{ newPath, files: [{ path, links }], failed }`, and now run off the main thread since they walk the workspace. Grant checks and destination logic are unchanged.
- **Link scan (`note.rs`, `md_links.rs`)**: link targets carry byte ranges, inline code spans are skipped (the index used to count `[[x]]` inside backticks, which the renderer never links), and relative markdown destinations are collected.
- **Index (`index.rs`)**: `apply_changes` expands a newly arrived folder into the notes under it and drops the notes under a folder that is gone, so a folder rename reported as its two paths re-indexes every note inside. A folder that already holds indexed notes is not re-walked on its own events.
- **Canvas (`canvas.rs`)**: `file_values` finds each `"file"` string token in the raw JSON so a card's path is spliced in place, keeping the board's formatting and key order.
- **Frontend flow (`useRelocation`)**: dry run, then the confirmation dialog (skipped when nothing links to the entry), then saves affected tabs holding unsaved edits, applies, re-points tabs, marks rewritten files as Glyph's own saves, reloads the open ones and refreshes the index so backlinks and the graph update immediately. A move confirmed after the workspace changed behind the prompt is dropped, and a tab edited while the move ran keeps what was typed. A failed write shows a persistent notice naming the file.
- **Dialogs**: `RelinkConfirmModal` lists files relative to the workspace with link counts and marks the ones whose unsaved changes will be saved first. `PromptModal` is the shared frame (focus on open, Tab trap, focus return, Escape cancels) and `UnsavedChangesModal` now uses it; `usePrompt` replaces `useUnsavedChangesPrompt` as the one promise bridge for both.
- **Reload**: `useDiskReload` is the tab reload `useTabEvents` used to hold inline. Its unsaved-edit guard (and, for a relink, a revision guard) runs against the latest state inside the state update.
- Locale strings in all five locales, `samples/README.md`, the README feature line, and the threat model row for `rename_path` / `move_path`.

Not covered: renames made outside Glyph, wikilinks to attachments (`![[img.png]]` is not indexed), wikilinks inside canvas text cards, indented code blocks, reference-style link definitions, and raw HTML `src`/`href`. The index still resolves canvas cards by name for graph edges, while the rewrite follows the renderer's board-relative reading.

## Risk classification

- [x] Persistence / data loss
- [ ] Secrets / credentials
- [ ] Network
- [x] Accessibility
- [ ] Bundle size / startup
- [ ] No risk areas touched

- **INV-1 (user edits are never silently discarded)**: affected tabs with unsaved edits are saved through the normal save path before the rewrite, and a failed save cancels the rename. A file changed on disk after planning keeps its new content. A tab dirty or edited while the move ran is not reloaded over. Evidence: `useTabs.mutations.test.tsx` "saves an affected tab before the rewrite, then reloads it and refreshes the index" and "does not rename when an affected tab cannot be saved"; `relink.rs` `a_file_changed_after_planning_keeps_its_new_content`; `useDiskReload.test.ts`.
- **INV-3 (stale results never win)**: the preview is only a list; the apply syncs the index and re-plans from disk. A confirmation that returns after the workspace changed is dropped. Rewritten tabs are matched through `movedPath`, so a reload lands on the right tab whether or not the re-point has rendered. Evidence: `vault/tests.rs` `a_dry_run_changes_nothing_and_a_rename_rewrites_and_reindexes`; `useTabs.mutations.test.tsx` "drops a rename confirmed after the workspace changed behind the prompt" and "reloads the moved note itself when its own links were rewritten"; `relink.rs` `a_file_edited_since_indexing_is_planned_from_disk`; `usePrompt.test.ts`.
- **INV-5 / INV-6 (external input untrusted, backend is the boundary)**: the commands keep their `ensure_workspace` and canonical-inside-root checks; every rewrite is written through `ensure_writable` and only to indexed notes and canvases inside the workspace. Markdown and canvas targets are resolved by a segment walk that never leaves the workspace, so a crafted card such as `//host/share/x.md` or `C:/x.md` is neither followed nor probed on disk. Evidence: `relink.rs` `a_card_naming_a_path_outside_the_workspace_is_never_followed` and `a_destination_outside_the_workspace_is_refused`; `commands/create.rs` `a_dry_run_is_denied_without_a_grant_and_otherwise_moves_nothing` plus the existing `..`, symlink and junction refusal tests.
- **INV-7 (partial results are explicit)**: a write that fails midway is a `failed` field naming the file, with the written files listed, and the UI shows a persistent notice. Evidence: `vault/tests.rs` `a_write_failing_midway_keeps_earlier_files_and_names_the_one_that_failed`; `useTabs.mutations.test.tsx` "names the file a rewrite stopped at and reads back only open files".
- **Untrusted rendering**: link parsing skips inline code to match the renderer, and the canvas scan never matches `"file"` inside a text card's escaped string. Evidence: `note.rs` `inline_code_holds_no_links`, `md_links.rs` tests, `canvas.rs` `file_values_are_found_by_key_and_never_inside_text`, `relink.rs` `code_is_never_rewritten`, `a_canvas_card_is_rewritten_in_place` and `a_card_is_relative_to_its_board_like_a_markdown_link`.
- **Accessibility**: the new dialog uses the shared `PromptModal` (labelled by title, described by message, focus trap, focus return, Escape cancels). Evidence: `PromptModal.test.tsx`, `RelinkConfirmModal.test.tsx`.

Acceptance-criteria link forms live in `relink.rs` tests: every wikilink form, the same-stem trap, a move that flips the same-directory tie-break, fences and inline code, markdown links, the moved note's own links, folder rename, canvas cards relative to their board, names a link cannot spell, and BOM/CRLF files.

## Testing

Automated gates run locally on Windows: `pnpm typecheck`, `pnpm check`, `pnpm test:coverage` (417 files, 3975 tests), `cargo test --lib` (852 tests), `cargo clippy --all-targets -- -D warnings`, and a patch coverage check of the changed lines on both sides. The rename and move flow was not exercised by hand in a running app.

- [ ] Tested on macOS
- [ ] Tested on Windows
- [ ] Tested on Linux

## Screenshots

None; the dialog reuses the existing unsaved-changes prompt styling.


hamidfzm:
Codecov's two remaining patch misses are `src-tauri/src/commands/create.rs` lines 127 and 242, the `#[tauri::command(async)]` attributes on `rename_path` and `move_path`. The async variant expands into IPC marshalling closures that llvm-cov attributes to the attribute line: one copy is inside the app's real `generate_handler!` in `lib.rs`, which no unit test can invoke, and the copy in the test build still reports zero hits even though `rename_and_move_answer_over_ipc` drives both commands over a real mock IPC call (success, a missing argument, and an ungranted root). Every line of the command bodies is covered.

The commands are async on purpose: each one now syncs the workspace index, so running them on the main thread would stall the UI on a large workspace. Both lines are left uncovered rather than dropping that.

The other two failing checks are unrelated to this change: `Cargo Audit` flags RUSTSEC-2026-0285 in `rustls` (no dependency changes here; a separate bump is in progress), and `Build / Android` fails in runner setup (`sdkmanager` cannot find package 'tools'), the same as other open PRs today.

